PR #29, opened 2014-06-29
Improve version generation at build time

dechamps:
This branch brings two improvements to how tinc determines its own version and build time:

- It makes sure the build date and time are regenerated every time tinc is built, even if there are no changes, so that the date and time returned by `tincd --version` is always accurate.
- It moves from hardcoded version numbers to autogenerated ones using `git describe` to ease maintenance and make `tincd --version` output more useful.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,6 +8,14 @@ ACLOCAL_AMFLAGS = -I m4
 
 EXTRA_DIST = COPYING.README README.android
 
+# If git describe works, force autoconf to run in order to make sure we have the
+# current version number from git in the resulting configure script.
+configure-version:
+	-git describe && autoconf --force
+
+# Triggering the README target means we are building a distribution (make dist).
+README: configure-version
+
 ChangeLog:
 	git log > ChangeLog
 
--- a/configure.ac
+++ b/configure.ac
@@ -1,10 +1,10 @@
 dnl Process this file with autoconf to produce a configure script.
 
 AC_PREREQ(2.61)
-AC_INIT([tinc], [1.1pre11])
+AC_INIT([tinc], m4_esyscmd([echo -n "`git describe || echo UNKNOWN | sed 's/release-//'`"]))
 AC_CONFIG_SRCDIR([src/tincd.c])
 AC_GNU_SOURCE
-AM_INIT_AUTOMAKE([check-news std-options subdir-objects -Wall])
+AM_INIT_AUTOMAKE([std-options subdir-objects -Wall])
 AC_CONFIG_HEADERS([config.h])
 
 # Enable GNU extensions.


dechamps:
@gsliepen: I'd like to remind you about this pull request. I would still like to get this in if we can make this work for you.

The first commit is useful because it allows the binaries to provide accurate information about the exact commit they were built from, which reduces confusion.

The second commit is useful to streamline the release process and reduce the risk of human error when making releases. It could be useful for you when you cut tinc releases, but it's also useful for me (and maybe other people) because I cut my own releases which I distribute privately to my VPN clients.

I've rebased the pull request and resolved the conflicts that crept up over time. Can you please take another look? This code works perfectly for me, and it's supposed to degrade gracefully when the git binary isn't available. I would be glad to help with any issues you might encounter while testing this.

gsliepen:
On Mon, May 04, 2015 at 01:53:48PM -0700, Etienne Dechamps wrote:

> @gsliepen: I'd like to remind you about this pull request. I would still like to get this in if we can make this work for you.
> 
> The first commit is useful because it allows the binaries to provide accurate information about the exact commit they were built from, which reduces confusion.
> 
> The second commit is useful to streamline the release process and reduce the risk of human error when making releases. It could be useful for you when you cut tinc releases, but it's also useful for me (and maybe other people) because I cut my own releases which I distribute privately to my VPN clients.
> 
> I've rebased the pull request and resolved the conflicts that crept up over time. Can you please take another look? This code works perfectly for me, and it's supposed to degrade gracefully when the git binary isn't available. I would be glad to help with any issues you might encounter while testing this.

There are still some issues:

- make dist doesn't work anymore if I just want to run that without
  doing a proper release, because autoconf puts in a check to see
  whether the version from configure.ac matches anything in the top of
  NEWS.

- People who do want to run autoreconf, but don't have git or are
  running on Windows, have a problem now.

- version_git.h should not end up in the tarball, I think adding a
  nodist_SOURCES = version_git.h should be enough to get rid of it.

- As a consequence of me prefixing release tags with "release-", things
  start looking ugly. Like the NEWS file.

I use make dist a lot for non-release versions for doing tests on
various machines. You can remove check-news from AM_INIT_AUTOMAKE().
Maybe a manual check can then be written in the release rule in
Makefile.am.

Another issue is that version information is kept correct in the
compiled version.c, but if you do a new commmit, make dist does
automatically put the new git version in the filename of the tarball.
That only happens if the configure script is regenerated. You could do
some dist-hook gymnastics, but it will be hard to do that cleanly and in
a portable way.

I actually don't use "make release" myself. If you don't care too much
about the filename of the dist tarballs, then what's left is the version
string in the binary. That's working fine (might be made prettier by
stripping "release-"). So if you make it so it only does this, I would
be perfectly happy.

-- 

dechamps:
> make dist doesn't work anymore if I just want to run that without doing a proper release, because autoconf puts in a check to see whether the version from configure.ac matches anything in the top of   NEWS.
> You can remove check-news from AM_INIT_AUTOMAKE().

Very well, I'll do that then. Presumably, if you're going to run `make dist` on something that's not a release tag, the check doesn't really make sense anyway.

> People who do want to run autoreconf, but don't have git or are running on Windows, have a problem now.

If you're running autoreconf on Windows it's reasonable to expect that git is available as well. Presumably there are two possible scenarios: either you're getting tinc from git, in which case you have to run autoreconf but that's fine since you had to use git to get the code in the first place, or you got the code from a tarball, in which case autoreconf has already been run for you. So in both these scenarios, you're fine.

It's true that if you find yourself in a situation in which you have to run autoreconf but git isn't available, autoreconf will fail. I will change the code slightly so that if git isn't available `configure.ac` will gracefully degrade by using "unknown" as the version number or something.

> version_git.h should not end up in the tarball, I think adding a nodist_SOURCES = version_git.h should be enough to get rid of it.

Sure. `nodist_SOURCES` doesn't seem to work, I'll take a closer look when I have more time.

> As a consequence of me prefixing release tags with "release-", things start looking ugly. Like the NEWS file.

That's a non-issue if we disable `check-news`.

> Another issue is that version information is kept correct in the compiled version.c, but if you do a new commmit, make dist does automatically put the new git version in the filename of the tarball. That only happens if the configure script is regenerated.

Yes, that's why I call `autoconf --force` from the `release` rule. I didn't consider the case where `dist` was called directly, since I assumed you were using `release`. I assumed wrong.

> I actually don't use "make release" myself.

How about we remove it then? People are still free to use their own release workflows in their own scripts.

> I actually don't use "make release" myself. If you don't care too much about the filename of the dist tarballs, then what's left is the version string in the binary. So if you make it so it only does this, I would be perfectly happy.

That's precisely what the first commit does, so feel free to merge that as soon as I get the `version_git.h` out of the tarball, while I figure out a cleaner solution for the `configure.ac` and `make dist` stuff (if I ever do).

gsliepen:
On Tue, May 05, 2015 at 01:04:30PM -0700, Etienne Dechamps wrote:

> That's precisely what the first commit does, so feel free to merge that as soon as I get the `version_git.h` out of the tarball, while I figure out a cleaner solution for the `configure.ac` and `make dist` stuff (if I ever do).

Ok, I found out how to get rid of version_git.h from the tarball, just
remove it from tinc_SOURCES and tincd_SOURCES.

I changed your way of generating version_git.h somewhat, because at
least on my system, even if git describe fails, the assignment
GIT_DESCRIPTION=`git describe` still succeeds, and therefore the &&
doesn't catch the error. So I just put git describe && ... as the very
first thing.

-- 

dechamps:
@gsliepen: I changed the autoconf commit. The new code should (hopefully) address your concerns. As far as I can tell, the only regressions that are introduced are:

- If you run `autoconf` outside of a git repo, or without a working `git` command, you will end up with `UNKNOWN` as the version number.
- Running `make dist` is slower since it has to run `autoconf` and `configure`.

Feel free to merge if these regressions are acceptable to you.

gsliepen:
On Sat, May 09, 2015 at 04:15:37AM -0700, Etienne Dechamps wrote:

> @gsliepen: I changed the autoconf commit. The new code should (hopefully) address your concerns. As far as I can tell, the only regressions that are introduced are:
> 
> - If you run `autoconf` outside of a git repo, or without a working `git` command, you will end up with `UNKNOWN` as the version number.

That's fine. I might replace UNKNOWN with an explicit version number
before a release. On the other hand I don't care too about it.

> - Running `make dist` is slower since it has to run `autoconf` and `configure`.

That's fine.

> Feel free to merge if these regressions are acceptable to you.

Done, although it had an error causing it not to remove the "release-"
part of the tag; | has precedence over || in shell scripts, so the sed
was only applied to the echo UNKNOWN. Parentheses came to the rescue.
Also, m4_esyscmd_s() conveniently removes the end-of-line, so no need
for the echo -n "`...`" hack :)

-- 

PR #45, opened 2014-07-30
Send a keep alive signal every second to keep UDP tunnels from collapsin...

vollkorn1982:
...g

Some routes might drop UDP tunnels through NAT firewalls after a very
short time frame. This leads to data being sent to the corresponding port
being lost. To prevent this every second a small mtu probe packet is sent
in order to keep the tunnel open and working. This does not eliminate the
possibility of packet loss due to eager routers clearing their tables, but
reduces the packet loss significantly.

### DIFF
--- a/src/net_packet.c
+++ b/src/net_packet.c
@@ -129,7 +129,6 @@ static void send_mtu_probe_handler(void *data)
 	{
 		logger(DEBUG_TRAFFIC, LOG_INFO, "Trying to send MTU probe to unreachable or rekeying node %s (%s)", n->name, n->hostname);
 		n->mtuprobes = 0;
-		logger(DEBUG_TRAFFIC, LOG_INFO, "state_machine: start rcv1;   %s (%s)", n->name, n->hostname);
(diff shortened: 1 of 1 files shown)


dechamps:
I've filed #61. As you can imagine, it goes in a completely different direction compared to your pull request.

Unfortunately, sending keepalives was not part of my motivations when writing #61 - in fact, it probably makes it worse in that regard. That said, if you set `UDPDiscoveryTimeout` from my pull request to an aggressive value, it will force tinc to assume that a UDP tunnel that has not seen any activity in the last few seconds should not be relied upon. That should fix your problem, as long as the inactivity timeout on your NAT is not set crazy low.

mark-stopka:
So should this be addressed further, or closed as non-relevant anymore?

vollkorn1982:
I did not test if the code in #61 does help this problem and won't do so anymore, since I'm not working with tinc anymore. But if the code does what @dechamps says it would do, then I think the problem is solved since the original problem was lost communication when connections dropped out.

PR #61, opened 2015-01-02
MTU discovery improvements

dechamps:
As promised, here is the result of my work on tinc's MTU discovery code.

This massive pull request has three major goals, which are described below.

# Code simplification

The current MTU discovery code is quite difficult to understand. The crux of the problem is the ` send_mtu_probe_handler()` function which implements a complicated and hard to read state machine.

I believe the main reason for the complexity if because the MTU discovery code is trying to do too many things at once: it's trying to estimate the MTU for the link, while making sure UDP connectivity is still up and running. These are two separate concerns intermingled in the same code, which increases state machine complexity.

In this pull request, I separate these two concerns by introducing a new concept, "UDP discovery". UDP discovery concentrates on keeping the UDP tunnel alive and checking its status, while MTU discovery concentrates on improving and fixing tunnel MTU as soon as UDP discovery establishes a working tunnel. This massively simplifies the MTU discovery code, which as a result only has two states to care about: initial discovery and "steady-state" (i.e. maxmtu checking). Separating these two concerns also makes it easier to implement the next improvements described below.

# Less network chatter

Something that's been grinding my gears for quite some time is tinc's eagerness to send probes when the original reason for sending them is gone. Consider a scenario in which a node makes a DNS request to another node over the VPN, and then the link between these two goes silent. From the high-level VPN point of view, that's one single packet going over the wire in each direction. In response to this, tinc will start MTU discovery, which in some cases can result in 90 packets (~126KB) going over the wire. Then, it will continue maintaining the tunnel basically *forever*, sending 3 packets (~4KB) every few seconds. This is in response to sending *one single small* VPN packet. The resulting write amplification is somewhat absurd.

While this is not a critical problem per se, it results in lots of network pollution, and makes tinc look "chatty" and "verbose". The inefficiency is detrimental to mobile devices (battery time) and constrained network links. More importantly, it does not scale: if a tinc node sends one packet to 1000 other nodes and then goes silent (e.g. P2P application), the resulting chatter will look truly awful.

The solution implemented in this pull request is simple: if you're not sending any real VPN packets, then you don't get to send any probes. Probes can only be sent from the `send_packet()` TX path. No packets, no probes. This results in the following behavior:

- If a node sends one packet and then goes silent, one single probe might be sent (well, technically, maybe two, if both UDP discovery and MTU discovery want one). That's it. After that, the physical link goes completely silent.
- If a node sends a slow trickle of packets, UDP/MTU discovery won't go faster than the packet rate. This way, we ensure we're not sending inappropriate amounts of control traffic compared to real traffic.
- If a node sends packets at a high rate (e.g. downloading), then UDP/MTU discovery behaves like it currently does - it will go at full speed. Indeed, if the tunnel is heavily used, it makes sense to optimize it as much as we can.
- If a link goes silent for a while (`UDPDiscoveryTimeout`), tinc can't assume the UDP link is usable anymore. Therefore it will reset UDP discovery and MTU discovery information. If the link gets used again, the first packet will go over TCP (which is fine), while tinc re-establishes a UDP tunnel and rediscovers the MTU. The latter operation is not really that expensive thanks to the next improvement below.

The keyword here is *adaptive*. tinc adapts to the real usage of the tunnel between two nodes, and will send probes more or less aggressively depending on how much the tunnel is used.

# Improved MTU discovery algorithm

This pull request also contains two changes to the core of the MTU discovery algorithm itself, yielding massive improvements with regard to convergence time and accuracy.

The first change makes the MTU probing code send one packet at a time instead of three. The frequency is adjusted from 1 second to 0.3 second to compensate, so that the number of packets per second stays the same. The rationale is that it leads to more efficient feedback, as the size of each probe is based on the previous replies; therefore, allowing replies to come in *between* probes decreases the total number of probe packets required to converge.

The second change moves from a naive random probe size model to a more sophisticated mathematical model with the goal of more efficiently covering the remaining MTU interval. This leads to massive improvements in convergence time and accuracy.

With these two improvements, tinc is typically able to converge on the precise (byte-level accuracy) MTU in less than 10 probe packets (3 seconds!). In contrast, the current random algorithm gives highly variable results, sometimes using *6 times* more probe packets to even get remotely close to the actual MTU.

This pull request supersedes #45.


gsliepen:
On Fri, Jan 02, 2015 at 02:46:24AM -0800, Etienne Dechamps wrote:

> As promised, here is the result of my work on tinc's MTU discovery code.
> 
> This massive pull request has three major goals, which are described below.
> 
> # Code simplification
> 
> The current MTU discovery code is quite difficult to understand. The crux of the problem is the ` send_mtu_probe_handler()` function which implements a complicated and hard to read state machine.

I completely agree.

> I believe the main reason for the complexity if because the MTU discovery code is trying to do too many things at once: it's trying to estimate the MTU for the link, while making sure UDP connectivity is still up and running. These are two separate concerns intermingled in the same code, which increases state machine complexity.

Well, in my mind it's one and the same thing; it's just trying to
determine a minimum value for the MTU, so that everything up to that
size can be sent via UDP, the rest goes via TCP. But yes, the state
machine is ugly.

> In this pull request, I separate these two concerns by introducing a new concept, "UDP discovery". UDP discovery concentrates on keeping the UDP tunnel alive and checking its status, while MTU discovery concentrates on improving and fixing tunnel MTU as soon as UDP discovery establishes a working tunnel. This massively simplifies the MTU discovery code, which as a result only has two states to care about: initial discovery and "steady-state" (i.e. maxmtu checking). Separating these two concerns also makes it easier to implement the next improvements described below.

I'm not against this :)

> # Less network chatter
> 
> Something that's been grinding my gears for quite some time is tinc's eagerness to send probes when the original reason for sending them is gone. Consider a scenario in which a node makes a DNS request to another node over the VPN, and then the link between these two goes silent. From the high-level VPN point of view, that's one single packet going over the wire in each direction. In response to this, tinc will start MTU discovery, which in some cases can result in 90 packets (~126KB) going over the wire. Then, it will continue maintaining the tunnel basically *forever*, sending 3 packets (~4KB) every few seconds. This is in response to sending *one single small* VPN packet. The resulting write amplification is somewhat absurd.

That's indeed a waste of bandwidth.

> While this is not a critical problem per se, it results in lots of network pollution, and makes tinc look "chatty" and "verbose". The inefficiency is detrimental to mobile devices (battery time) and constrained network links. More importantly, it does not scale: if a tinc node sends one packet to 1000 other nodes and then goes silent (e.g. P2P application), the resulting chatter will look truly awful.
> 
> The solution implemented in this pull request is simple: if you're not sending any real VPN packets, then you don't get to send any probes. Probes can only be sent from the `send_packet()` TX path. No packets, no probes.

That's an excellent solution to the problem!

> This results in the following behavior:
> 
> - If a node sends one packet and then goes silent, one single probe might be sent (well, technically, maybe two, if both UDP discovery and MTU discovery want one). That's it. After that, the physical link goes completely silent.

Great.

> - If a node sends a slow trickle of packets, UDP/MTU discovery won't go faster than the packet rate. This way, we ensure we're not sending inappropriate amounts of control traffic compared to real traffic.
> - If a node sends packets at a high rate (e.g. downloading), then UDP/MTU discovery behaves like it currently does - it will go at full speed. Indeed, if the tunnel is heavily used, it makes sense to optimize it as much as we can.
> - If a link goes silent for a while (`UDPDiscoveryTimeout`), tinc can't assume the UDP link is usable anymore. Therefore it will reset UDP discovery and MTU discovery information. If the link gets used again, the first packet will go over TCP (which is fine), while tinc re-establishes a UDP tunnel and rediscovers the MTU. The latter operation is not really that expensive thanks to the next improvement below.

There is one issue here, and that is UDP hole punching. If you have
three nodes, A B and C, and A and B are behind a NAT and C not, then
they need C to help them punch holes. This can only happen if C has
active UDP connections with A and B. This is actually the reason that
tinc tried to keep UDP alive for every meta-connection. There's some
limitations in the 1.0.x protocol that prevent this from being done on
demand only.

> The keyword here is *adaptive*. tinc adapts to the real usage of the tunnel between two nodes, and will send probes more or less aggressively depending on how much the tunnel is used.
> 
> # Improved MTU discovery algorithm
> 
> This pull request also contains two changes to the core of the MTU discovery algorithm itself, yielding massive improvements with regard to convergence time and accuracy.
> 
> The first change makes the MTU probing code send one packet at a time instead of three. The frequency is adjusted from 1 second to 0.3 second to compensate, so that the number of packets per second stays the same. The rationale is that it leads to more efficient feedback, as the size of each probe is based on the previous replies; therefore, allowing replies to come in *between* probes decreases the total number of probe packets required to converge.

Yes, however the burst bandwidth estimation code then indeed does not
work anymore. Although it might be inaccurate, it does at least get the
order of magnitude correct. On the other hand, it's not used for
anything at the moment.

> The second change moves from a naive random probe size model to a more sophisticated mathematical model with the goal of more efficiently covering the remaining MTU interval. This leads to massive improvements in convergence time and accuracy.

I'm very sceptical about this. It's very easy to optimize the algorithm
for YOUR network, it's something else to make an algorithm that works
well on ANY network. I did some simulations myself in the past with
different algorithms, and calculated the mean/median/sdev of them for
ALL possible path MTUs, with and without ICMP Packet Too Big responses.
The random sized packets algorithm median number of packets to fix the
MTU sees much less peaks when you vary the actual PMTU than other
algorithms, and on average over all possible PMTUs it is performing more
or less just as well as deterministic algorithms. I'll try to run the
simulations again with your algorithm.

> With these two improvements, tinc is typically able to converge on the precise (byte-level accuracy) MTU in less than 10 probe packets (3 seconds!). In contrast, the current random algorithm gives highly variable results, sometimes using *6 times* more probe packets to even get remotely close to the actual MTU.

Well, before tinc usually fixed the MTU in 7 probes if ICMP Packet Too
Big responses are not filtered, and just switching to 1 packet every 0.3
seconds would improve that. But 3 seconds is indeed a very nice result.

> You can merge this Pull Request by running:
> 
>   git pull https://github.com/dechamps/tinc pmtu

I tried it out, but there are some issues. Running it between two nodes
on a LAN in switch mode, I see that the broadcast packets caused by IPv6
auto IP assigment and Avahi when the interface goes up don't trigger UDP
probes at all. Only when there are no broadcast packets for > 10
seconds, and I then try to ping the other side, will UDP probes be sent.

After the first UDP probe is succesfully received, MTU probing starts,
and I see this:

Using system-provided maximum tinc MTU for xar (10.1.1.1 port 28540):
1472
Sending UDP probe length 1472 to xar (10.1.1.1 port 28540)
Sending UDP probe length 1471 to xar (10.1.1.1 port 28540)
Sending UDP probe length 1470 to xar (10.1.1.1 port 28540)
[...]
Sending UDP probe length 1460 to xar (10.1.1.1 port 28540)
Sending UDP probe length 1459 to xar (10.1.1.1 port 28540)
Got type 2 UDP probe reply 1459 from xar (10.1.1.1 port 28540)
Fixing MTU of xar (10.1.1.1 port 28540) to 1459 after 1 probes

This is because choose_initial_maxmtu() doesn't apply the overhead of
legacy UDP packet encryption, which can vary depending on Cipher, Digest
and MACLength. In this very case, it's sort of nice it counts down until
it reaches a size that works, but I think that it should only retry if
n->mtuprobes >= 0 and should increase the counter inside the for loop.

Apart from the algorithm for choosing packet sizes, I think the MTU
probing code is very nice. As for UDP probing, I would make a few
changes. First of all, while status.udp_confirmed is false, instead of
using udp_discovery_value, a smaller value like 0.3 seconds could be used.
Also, for nodes that we also have a meta-connection with, try_udp()
should be called every udp_discovery_interval to keep the UDP mappings alive.
Maybe this can be done just by looping over the connection_list and
calling try_udp(c->node) in periodic_handler() in net.c.

Do you want to change anything before I pull your changes? I don't mind
pulling now and working on the details later.

-- 

dechamps:
> There is one issue here, and that is UDP hole punching. If you have three nodes, A B and C, and A and B are behind a NAT and C not, then they need C to help them punch holes. This can only happen if C has active UDP connections with A and B. This is actually the reason that tinc tried to keep UDP alive for every meta-connection. There's some limitations in the 1.0.x protocol that prevent this from being done on demand only.

Oh, right, I always wondered why tinc viewed direct neighbors as "special snowflakes" in that regard, that answers it. As you probably already noticed, this is not an issue with SPTPS because it will spontaneously relay in that case, so indirect UDP tunnels will always get established if necessary. I understand that the non-SPTPS path doesn't do that (it doesn't know how to fall back to *indirect* UDP), hence the problem.

To be honest, I'm not sure how "well" you want 1.0.x-1.1.x cohabitation to work. I always thought this was simply to ease the transition path, and that no-one would keep an heterogeneous graph in the long term, and that is was therefore acceptable to have less-than-optimal performance characteristics between 1.0.x and 1.1.x nodes, as long as communication is still possible (e.g. via TCP). That said this looks like an easy fix, so we might as well do that.

> Yes, however the burst bandwidth estimation code then indeed does not work anymore. Although it might be inaccurate, it does at least get the order of magnitude correct. On the other hand, it's not used for anything at the moment.

That's collateral damage I'm afraid. I strongly believe it's worth it, but it's your call of course :)

That said, you are quite lucky to at least get the correct order of magnitude. In my experience, if the link is just a little unpredictable (jitter, congestion), that thing returns completely absurd numbers.

> I'm very sceptical about this. It's very easy to optimize the algorithm for YOUR network, it's something else to make an algorithm that works well on ANY network.

Well, the algorithm I wrote is optimized for a MTU in the 1400-1450 range (which is typical of tinc running over typical physical networks), with no help from the system (i.e. no `EMSGSIZE`). It's not optimized for "my" network (which does have reliable `EMSGSIZE` notifications). It's optimized in a spreadsheet :) 

> I did some simulations myself in the past with different algorithms, and calculated the mean/median/sdev of them for ALL possible path MTUs, with and without ICMP Packet Too Big responses. The random sized packets algorithm median number of packets to fix the MTU sees much less peaks when you vary the actual PMTU than other algorithms, and on average over all possible PMTUs it is performing more or less just as well as deterministic algorithms.

Oh yeah, I spent a lot of time doing tons of simulations as well. Yes, I agree, it's surprisingly hard to beat the random algorithm. None of the formulas I tried managed to beat it, except the one I implemented in this pull request. It was the only one to show significantly better results compared to the random one. [Here's the spreadsheet that I used to determine that][sim]. I believe you should be able to copy it to your own Drive and then put in new numbers from there.

My main problem with the random algorithm is, well, it's random. In my opinion, that alone is a disadvantage, as by definition is leads to somewhat unpredictable results. Having an MTU that depends on the phase of the moon is not great. Of course there is always some nondeterminism in the form of packet loss, but that's not as bad.

In addition, my algorithm pretty much always guarantee it will get to byte-level accuracy in a reasonable amount of time (typically < 20 probes). With the random algorithm there are pathological cases where that will take a very long time (consider minmtu 1418, maxmtu 1518 and actual MTU 1419, each probe has a 1% chance of landing on 1419). That cannot happen with the one I implemented.

> Well, before tinc usually fixed the MTU in 7 probes if ICMP Packet Too Big responses are not filtered

All the numbers I mentioned assume a worst case scenario where these responses are filtered.

> Running it between two nodes on a LAN in switch mode, I see that the broadcast packets caused by IPv6 auto IP assigment and Avahi when the interface goes up don't trigger UDP probes at all. Only when there are no broadcast packets for > 10 seconds, and I then try to ping the other side, will UDP probes be sent.

That's weird. I think I've tested broadcast packets and they seemed to work correctly. I'll look into it.

> This is because choose_initial_maxmtu() doesn't apply the overhead of legacy UDP packet encryption, which can vary depending on Cipher, Digest and MACLength.

Yes. I had no idea how to calculate overhead for 1.0.x, so I cheated and applied my "we don't care that much about 1.0" philosophy instead :) For SPTPS packets the overhead calculation is accurate. If you know how to calculate that for 1.0, then by all means, be my guest.

> In this very case, it's sort of nice it counts down until it reaches a size that works, but I think that it should only retry if n->mtuprobes >= 0 and should increase the counter inside the for loop. 

Maybe. I don't really see that as a problem, but I'm fine either way.

> As for UDP probing, I would make a few changes. First of all, while status.udp_confirmed is false, instead of using udp_discovery_value, a smaller value like 0.3 seconds could be used.

I guess that makes sense, probes sent while udp_confirmed is false are very lightweight anyway (16 bytes payload). I'll change the pull request.

> Also, for nodes that we also have a meta-connection with, try_udp() should be called every udp_discovery_interval to keep the UDP mappings alive. Maybe this can be done just by looping over the connection_list and calling try_udp(c->node) in periodic_handler() in net.c.

That's related to the hole punching issue with 1.0.x that you mentioned above, right? In that case I'm fine with you making the change, you seem to care more about 1.0.x than I do :)

> Do you want to change anything before I pull your changes?

I'll try to investigate the problem with broadcasts and fix it. I'll also make the change to fine-tune the UDP discovery interval. After that, feel free to merge and do anything you want with regard to the legacy stuff and the handling of maxmtu changes in `try_mtu()`.

  [sim]: https://docs.google.com/spreadsheets/d/1r1RaKBIFAw6l-5GIQjHv_dBPRRq-JAqUHnCFWR3rW_A/edit?usp=sharing

gsliepen:
On Fri, Jan 02, 2015 at 10:07:23AM -0800, Etienne Dechamps wrote:

> > There is one issue here, and that is UDP hole punching. If you have three nodes, A B and C, and A and B are behind a NAT and C not, then they need C to help them punch holes. This can only happen if C has active UDP connections with A and B. This is actually the reason that tinc tried to keep UDP alive for every meta-connection. There's some limitations in the 1.0.x protocol that prevent this from being done on demand only.
> 
> Oh, right, I always wondered why tinc viewed direct neighbors as "special snowflakes" in that regard, that answers it. As you probably already noticed, this is not an issue with SPTPS because it will spontaneously relay in that case, so indirect UDP tunnels will always get established if necessary. I understand that the non-SPTPS path doesn't do that (it doesn't know how to fall back to *indirect* UDP), hence the problem.

I don't really see how this is different for SPTPS... the issue is that
node C needs to inform A and B of their "reflexive" UDP address+port.
This is done only when A sends an ANS_KEY to B via C, and vice versa.
And that only happens during the initial SPTPS handshake between A and
B. I know, this is stupid. But that's because all this is shoe-horned
into a protocol that even tinc 1.0.0 nodes should still be able to work
with. I'll think a bit about how to improve this, at least between 1.1
nodes this can be fixed.

> To be honest, I'm not sure how "well" you want 1.0.x-1.1.x cohabitation to work. I always thought this was simply to ease the transition path, and that no-one would keep an heterogeneous graph in the long term,

I hope things are better at your work, but at my last job as a UNIX
sysadmin we had a translation table that had examples like this:

"On paper: legacy system will be phased out in three months -> In
reality: system has to be supported until the heat death of the
universe."

Any network with >10 devices that are not under the control of a single
individual is bound to be very heterogenous.

> and that is was therefore acceptable to have less-than-optimal performance characteristics between 1.0.x and 1.1.x nodes, as long as communication is still possible (e.g. via TCP). That said this looks like an easy fix, so we might as well do that.

I would like to keep performance of mixed networks as optimal as
possible, only sacrificing it if it would be too burdensome.

> > Yes, however the burst bandwidth estimation code [...], you are quite lucky to at least get the correct order of magnitude. In my experience, if the link is just a little unpredictable (jitter, congestion), that thing returns completely absurd numbers.

Ok. Well it just gave the raw values from a single burst, if it were
used for anything it should have been smoothed in some way. Did you get
numbers substantially higher than the actual bandwidth?

> > I did some simulations myself in the past with different algorithms, and calculated the mean/median/sdev of them for ALL possible path MTUs, with and without ICMP Packet Too Big responses. The random sized packets algorithm median number of packets to fix the MTU sees much less peaks when you vary the actual PMTU than other algorithms, and on average over all possible PMTUs it is performing more or less just as well as deterministic algorithms.
> 
> Oh yeah, I spent a lot of time doing tons of simulations as well. Yes, I agree, it's surprisingly hard to beat the random algorithm. None of the formulas I tried managed to beat it, except the one I implemented in this pull request. It was the only one to show significantly better results compared to the random one. [Here's the spreadsheet that I used to determine that][sim]. I believe you should be able to copy it to your own Drive and then put in new numbers from there.

I have several drives, but they are all in my own computers ;)

In any case, I have put your algorithm into my own simulation, and there
is a surprising result. The random algorithm, modified to not do burst
of 3 packets anymore but just 1 packet every 0.3 seconds, is on average
better than your algorithm by a factor 1.5, but only if sendto() can
fail with EMSGSIZE. If not, then your algorithm finds the actual value
of the MTU in only a quarter of the number of probes (I'm not counting
any extra probes because it doesn't know maxmtu yet here). 

Your algorithm has some weird behavior. It is quite sensitive to the
multiplier. Indeed, 0.982 optimizes it for the range 1410-1460, and if
you set it to 1 it actually performs quite a bit worse at MTUs >1400. 
Also, you would expect that the behavior improves with EMSGSIZE, but
for your algorithm it just gives different results, sometimes better,
sometimes worse.

Your algorithm is slightly more sensitive to packet loss, but this only
becomes an issue when it's more than 20%. That's actually a very good
result for a completely deterministic algorithm.

I'll try to see if I can combine the best of both algorithms :)

> In addition, my algorithm pretty much always guarantee it will get to byte-level accuracy in a reasonable amount of time (typically < 20 probes). With the random algorithm there are pathological cases where that will take a very long time (consider minmtu 1418, maxmtu 1518 and actual MTU 1419, each probe has a 1% chance of landing on 1419). That cannot happen with the one I implemented.

That's indeed true, and probably that is why your algorithm converges so
much faster than the random one. That got me thinking, it might be nice to
keep some extra state, namely what size packet was sent last time. If no
response is received, then randomly/deterministically choose to either
send a packet that is halfway between that size and minmtu, or send one
between maxmtu and minmtu. In the latter case, don't update the "last sent
packet size). I modified my algorithm in the simulation, and with some
experimentation found that the best results are when 1 in 8 times I
always try a random packet size between minmtu and maxmtu, the rest of
the times I choose minmtu + 1 + (lastsize - minmtu) / 3. The 1 in 8
times matches your probes_per_cycle, and the reduction in size by 3
matches your powf(). The resulting algorithm has similar performance as
yours without EMSGSIZE, and somewhat better performance with EMSGSIZE.

It's too late now, but tomorrow I'll send you the simulation code.

> > This is because choose_initial_maxmtu() doesn't apply the overhead of legacy UDP packet encryption, which can vary depending on Cipher, Digest and MACLength.
> 
> Yes. I had no idea how to calculate overhead for 1.0.x, so I cheated and applied my "we don't care that much about 1.0" philosophy instead :) For SPTPS packets the overhead calculation is accurate. If you know how to calculate that for 1.0, then by all means, be my guest.

Sure, I'll add it :)

> > Also, for nodes that we also have a meta-connection with, try_udp() should be called every udp_discovery_interval to keep the UDP mappings alive. Maybe this can be done just by looping over the connection_list and calling try_udp(c->node) in periodic_handler() in net.c.
> 
> That's related to the hole punching issue with 1.0.x that you mentioned above, right? In that case I'm fine with you making the change, you seem to care more about 1.0.x than I do :)

Ok :)

> > Do you want to change anything before I pull your changes?
> 
> I'll try to investigate the problem with broadcasts and fix it. I'll also make the change to fine-tune the UDP discovery interval. After that, feel free to merge and do anything you want with regard to the legacy stuff and the handling of maxmtu changes in `try_mtu()`.

Ok.

-- 

dechamps:
> I don't really see how this is different for SPTPS... the issue is that node C needs to inform A and B of their "reflexive" UDP address+port. This is done only when A sends an ANS_KEY to B via C, and vice versa. And that only happens during the initial SPTPS handshake between A and B. I know, this is stupid. But that's because all this is shoe-horned into a protocol that even tinc 1.0.0 nodes should still be able to work with. I'll think a bit about how to improve this, at least between 1.1 nodes this can be fixed.

Ah, yes, of course. That makes sense. Actually I knew about this at some point in the past, because I remember noticing the potential race condition issues related to piggybacking on this message: if `ANS_KEY` is sent before the intermediate nodes have a full picture about their UDP situation, then it's ineffective, and it won't be resent afterwards.

Well, the next thing I was planning to do after this was to implement [UDP info messages][udpinfo], which is designed specifically to address this problem. It will make timing irrelevant, because these messages will be sent periodically so that UDP address information can be updated in real-time as intermediate nodes establish tunnels (which they always do because SPTPS relaying fallback). Of course that will only work if every node in the path is able to understand these messages, so that excludes 1.0.x. Is that okay with you? From your message I get the feeling you had something similar in mind.

> In any case, I have put your algorithm into my own simulation, and there is a surprising result. The random algorithm, modified to not do burst of 3 packets anymore but just 1 packet every 0.3 seconds, is on average better than your algorithm by a factor 1.5, but only if sendto() can fail with EMSGSIZE. If not, then your algorithm finds the actual value of the MTU in only a quarter of the number of probes (I'm not counting any extra probes because it doesn't know maxmtu yet here). 
> Also, you would expect that the behavior improves with EMSGSIZE, but for your algorithm it just gives different results, sometimes better, sometimes worse.

Well, I fine-tuned it for the worst case scenario (no `EMSGSIZE`). Maybe by tweaking it a little bit more we can get to a compromise between the best case scenario and the worst case scenario.

> Your algorithm has some weird behavior. It is quite sensitive to the multiplier. Indeed, 0.982 optimizes it for the range 1410-1460, and if you set it to 1 it actually performs quite a bit worse at MTUs >1400.

Setting it to 1 is a bad idea, because it means the first probe is always set to maxmtu (`offset = pow(interval, 1)`). If `EMSGSIZE` is not returned, it means the algorithm will send one 1518-byte probe at the beginning of each cycle, and that probe is unlikely to get back unless the physical network supports jumbo frames.

The reason why the multiplier is very sensitive is because it is directly multiplying an exponent. Exponentials are known to be quite sensitive :) 

> I'll try to see if I can combine the best of both algorithms :)

You might want to try using your random algorithm but with a non-uniform distribution biased towards small probes. The problem is, you're unlikely to get the nice behavior of trying large probes first (which is what my algorithm does) if you do it that way.

> That's indeed true, and probably that is why your algorithm converges so much faster than the random one. That got me thinking, it might be nice to keep some extra state, namely what size packet was sent last time.

I thought about it, but I was trying to refrain from adding more complexity and state to the algorithm, so I did not investigate that option. But if you're fine with it, by all means, be my guest.

  [udpinfo]: http://www.tinc-vpn.org/pipermail/tinc-devel/2014-September/000674.html

dechamps:
I've added a commit for the discovery timeout change. I've set it to 2 seconds, that seems more appropriate than 0.3 second which sounds very aggressive to me. Tunnel establishment getting delayed by two or four seconds because of lost packets doesn't seem like a big deal to me. Feel free to readjust if you disagree.

I tried to reproduce your issue with broadcasts, but was unable to. I've set up a testbed where I only send broadcasts and nothing else, and everything works normally. The alternative would have been very surprising because broadcasts use the same `send_packet()` TX path as everything else - there's no reason why they would act any different from normal packets as far as UDP discovery is concerned. Since I can't reproduce it, I'm afraid you're gonna have to try to debug it yourself. Is there anything non-standard about your test configuration?

I'm okay with merging now.

gsliepen:
On Sat, Jan 03, 2015 at 01:26:40AM -0800, Etienne Dechamps wrote:

> > I don't really see how this is different for SPTPS... the issue is that node C needs to inform A and B of their "reflexive" UDP address+port. This is done only when A sends an ANS_KEY to B via C, and vice versa. And that only happens during the initial SPTPS handshake between A and B. I know, this is stupid. But that's because all this is shoe-horned into a protocol that even tinc 1.0.0 nodes should still be able to work with. I'll think a bit about how to improve this, at least between 1.1 nodes this can be fixed.
> 
> Ah, yes, of course. That makes sense. Actually I knew about this at some point in the past, because I remember noticing the potential race condition issues related to piggybacking on this message: if `ANS_KEY` is sent before the intermediate nodes have a full picture about their UDP situation, then it's ineffective, and it won't be resent afterwards.
> 
> Well, the next thing I was planning to do after this was to implement [UDP info messages][udpinfo], which is designed specifically to address this problem. It will make timing irrelevant, because these messages will be sent periodically so that UDP address information can be updated in real-time as intermediate nodes establish tunnels (which they always do because SPTPS relaying fallback). Of course that will only work if every node in the path is able to understand these messages, so that excludes 1.0.x. Is that okay with you? From your message I get the feeling you had something similar in mind.

Yes. But between 1.1 and 1.0 nodes it should still be possible to just
keep the UDP mappings alive with small UDP probes. In fact, the 1.0
nodes will do this anyway... so maybe this doesn't require any effort.

> > I'll try to see if I can combine the best of both algorithms :)
> 
> You might want to try using your random algorithm but with a non-uniform distribution biased towards small probes. The problem is, you're unlikely to get the nice behavior of trying large probes first (which is what my algorithm does) if you do it that way.

That might be interesting. Anyway, I put a tarball with the simulation
code and the results online:

http://tinc-vpn.org/temp/simpmtu.tar.gz

I ran the simulations with both 1% and 10% packet loss. With 1% packet
loss, the result is virtually the same as 0% packet loss for all
algorithms. I also ran it with support for jumbograms (maximum MTU
9018).

-- 

gsliepen:
On Sat, Jan 03, 2015 at 02:24:57AM -0800, Etienne Dechamps wrote:

> I tried to reproduce your issue with broadcasts, but was unable to. I've set up a testbed where I only send broadcasts and nothing else, and everything works normally. The alternative would have been very surprising because broadcasts use the same `send_packet()` TX path as everything else - there's no reason why they would act any different from normal packets as far as UDP discovery is concerned. Since I can't reproduce it, I'm afraid you're gonna have to try to debug it yourself. Is there anything non-standard about your test configuration?

I didn't think so... but maybe I made a mistake in my setup, I'll try it
again.

> I've added a commit for the discovery timeout change. I've set it to 2 seconds, that seems more appropriate than 0.3 second which sounds very aggressive to me. Tunnel establishment getting delayed by two or four seconds because of lost packets doesn't seem like a big deal to me. Feel free to readjust if you disagree.

Well, since those probes are only sent when there is other traffic
anyway, I don't see why a shorter interval would hurt. If packet loss is
completely stochastic anyway, then the time between the probes should
not matter much: if you say two or four seconds, that just means 1 or 2
packets get lost. If you decrease the interval, it means tunnel
establishment only gets delayed 0.3 or 0.6 seconds. Especially when you
stop UDP probes much more agressively than tinc does now, you want it to
come back up faster as well IMHO.

> I'm okay with merging now.

Great!

-- 

dechamps:
> Well, since those probes are only sent when there is other traffic anyway, I don't see why a shorter interval would hurt.

Well, it hurts because it increases network chatter, though how much chatter is acceptable is a pure judgement call, so I'm fine with anything you think is best.

That said, I would urge you to consider the worst case scenario of sustained traffic from a node behind a TCP-only firewall: in that case 4 probes will be sent per interval (1 for the destination node, 1 for the relay, and both are in duplicate because of local discovery). If you do that every 0.3 second, you end up with 13 pings per second, which seem a little extreme to me.

gsliepen:
On Sat, Jan 03, 2015 at 07:25:49AM -0800, Etienne Dechamps wrote:

> > Well, since those probes are only sent when there is other traffic anyway, I don't see why a shorter interval would hurt.
> 
> Well, it hurts because it increases network chatter, though how much chatter is acceptable is a pure judgement call, so I'm fine with anything you think is best.
> 
> That said, I would urge you to consider the worst case scenario of sustained traffic from a node behind a TCP-only firewall: in that case 4 probes will be sent per interval (1 for the destination node, 1 for the relay, and both are in duplicate because of local discovery). If you do that every 0.3 second, you end up with 13 pings per second, which seem a little extreme to me.

Ah, maybe I forgot to mention, but I don't intend to do probes every 0.3
seconds all the time, but just for the first 10 probes or so when trying
to start up UDP. And if it fails just do it once every minute or so. I
have three intervals in my mind:

1) Short (0.3 s) during initial UDP probing, only for 10 packets max
2) Medium (10 s) while UDP traffic is flowing
3) Long (60 s) while we have traffic but status.udp_confirmed == false.

The rationale is to have quick (re-)establishment of UDP tunnels, then
keep it alive as long as it's in use, and if UDP doesn't work then only
try it again once in a while. Maybe 3) can be omitted and just use 10 s
as well, since we only send the packets when there is actual traffic
anyway.

-- 

dechamps:
> Ah, maybe I forgot to mention, but I don't intend to do probes every 0.3 seconds all the time, but just for the first 10 probes or so when trying to start up UDP. And if it fails just do it once every minute or so.

I don't think this is worth the extra complexity, but that's just my opinion.

Furthermore, if you're going to do something like this you might as well go all the way and implement exponential backoff, since that's best practice for this sort of thing as far as I know.

PR #65, opened 2015-03-08
Add UDP_INFO and MTU_INFO protocol messages

dechamps:
This pull request introduces two new protocol messages, which are protected by minor protocol versions 5 and 6, respectively:

- The `UDP_INFO` message is designed to make UDP hole punching (and other UDP trickery) more reliable, less prone to race conditions, and more reactive to changing network conditions. It will be especially useful when the `ANS_KEY` message will be deprecated in favor of native, fully SPTPS-aware alternatives.
- The `MTU_INFO` message is designed to solve the problem of finding the minimum MTU in an UDP SPTPS relay path.

Both messages follow a similar principle: the packet recipient sends the initial message from the RX path (honoring the principle of "no data traffic, no chatter") with a bogus set of initial values, and the message gets altered as it travels through the metagraph, with the values contained in the message becoming more and more relevant as it gets closer to its destination (i.e. the packet sender). The messages are sent on a regular interval, and do not elicit a response.


gsliepen:
On Sun, Mar 08, 2015 at 01:27:27PM -0700, Etienne Dechamps wrote:

> This pull request introduces two new protocol messages, which are protected by minor protocol versions 5 and 6, respectively:
> 
> - The `UDP_INFO` message is designed to make UDP hole punching (and other UDP trickery) more reliable, less prone to race conditions, and more reactive to changing network conditions. It will be especially useful when the `ANS_KEY` message will be deprecated in favor of native, fully SPTPS-aware alternatives.
> - The `MTU_INFO` message is designed to solve the problem of finding the minimum MTU in an UDP SPTPS relay path.
> 
> Both messages follow a similar principle: the packet recipient sends the initial message from the RX path (honoring the principle of "no data traffic, no chatter") with a bogus set of initial values, and the message gets altered as it travels through the metagraph, with the values contained in the message becoming more and more relevant as it gets closer to its destination (i.e. the packet sender). The messages are sent on a regular interval, and do not elicit a response.

It looks very interesting. I don't have too much time to review it
properly at the moment, so I'll leave this in the queue for a while
longer.

The UDP/MTU_INFO messages are now only sent via TCP. But what if they
could be included in the UDP probe messages as well, for those parts of
the path that can use UDP?

-- 

dechamps:
> The UDP/MTU_INFO messages are now only sent via TCP. But what if they could be included in the UDP probe messages as well, for those parts of the path that can use UDP?

I suppose that's possible, but I'm not sure it's worth it. These messages are very small, are only sent when there is traffic, and are throttled to once every five seconds (a default which could probably be increased without too much risk). In addition, they are only sent when the circumstances make sense (i.e. if there is a usable direct tunnel, neither UDP nor MTU info messages get sent). Finally, when sent over TCP, intermediate nodes can opportunistically learn this information as well. Arguably, when there is traffic these messages basically replace PING/PONG messages that the nodes would send to each other instead, so the increase in chatter is somewhat relative.

Adding UDP as an alternative mechanism for sending these messages would be adding unnecessary complexity, IMHO.

Furthermore, I suspect there is a gotcha there: in the case of a dynamic relay path, the path in one direction can be different from the relay path in the other direction. In fact there is a very common scenario where that can happen: picture A ↔ B ↔ C ↔ D, where every node can reach each other directly, except A ↔ D. In that case A's dynamic relay path to D is simply B, whereas D's dynamic relay path to A is C (not B). This is because the current code simply uses the nexthop as the dynamic relay (even if a closer node can be reached directly - maybe this logic could be improved - currently this is inspired by how relaying works in the legacy protocol). If the relay path is different in one direction, then it means UDP info messages (which are sent in the opposite direction of traffic) will not travel through the relevant relays. Unless of course we always use nexthop as the destination for these UDP messages, but in that case, why are we using UDP in the first place?

In any case, there's no reason this can't be added in the future if it turns out to be feasible and useful.

PR #66, opened 2015-03-13
Add a new --syslog option for tincd.

seehuhn:
This commit adds a new command line option for tincd which allows to
use tincd in non-detached mode with log messages still going to
syslog.  The motivation for this change is to ease use of tincd
in Docker containers.


seehuhn:
Thanks! Are also interested in a backport of this patch for the 1.0 series?  I have one ready for my own use, happy to send a pull request if you want me to.

PR #74, opened 2015-04-04
tinc-1.1 better RTT estimated weight computation.

thorkill:
I have this problem:
Node A - slow CPU/embedded device
Node D - laptop/"road warrior"

while connecting to D on LAN the estimated wieght was ~2000. I can connect to this node via direct connection which is OK and the RTT ~1ms. But the tool which generates a network graph shows the edge between A and D as very slow - which is not fully true. With this patch we compute c->estimated_weight few times taking the average.

Better solution would be to have the last timestamp of the protocol internal communication.


thorkill:
After testing with more nodes this method doesn't work as expected - will try another approach.

PR #76, opened 2015-04-11
Implements edge weight updates based on average RTT.

thorkill:


thorkill:
The code of ponyhof branch changed many times. I am closing the pull request cause it may be incompatible with the current patches.

PR #80, opened 2015-05-10
Improve SPTPS operation over TCP

dechamps:
This pull request brings two important features to the SPTPS protocol:

- tinc now uses the relay code when receiving a SPTPS packet in an extended `REQ_KEY` message. This allows SPTPS packets to "jump" from TCP to UDP along the relay path if possible.

- A new protocol message, `SPTPS_PACKET`, is introduced for highly efficient SPTPS packet transport over TCP.

Further details can be found in the commit messages.

I am pleased to announce that this is the very last series of patches required to bridge the efficiency gap between the old legacy protocol and the SPTPS protocol. Thanks to the relaying mechanisms and efficient TCP transport, there is no performance downside for using SPTPS anymore.

### DIFF
--- a/src/connection.h
+++ b/src/connection.h
@@ -97,6 +97,7 @@ typedef struct connection_t {
 	struct buffer_t outbuf;
 	io_t io;                        /* input/output event on this metadata connection */
 	int tcplen;                     /* length of incoming TCPpacket */
+	int sptpslen;			/* length of incoming SPTPS packet */
 	int allow_request;              /* defined if there's only one request possible */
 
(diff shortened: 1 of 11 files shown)


dechamps:
Note: to prevent any confusion, the commit about `sptps_receive_data()` comes before the commit about raw TCP SPTPS packet transport. Github is showing them in the wrong order because it shows them in the order I wrote them, not the actual commit order.

dechamps:
Rebased to fix conflicts introduced by the merge of #79.

PR #94, opened 2015-07-07
Prevent heap-use-after-free and make additional checks which prevent memory leak

thorkill:


gsliepen:
Fixed in commits bda16410075a63b147179c15aa75640decf17ec7 and 3bf3d7d3e7d51034bda873861c52578f3abe6b5d. Thanks!

PR #96, opened 2015-09-03
Fix DecrementTTL option.

VittGam:
Hello,

This patch fixes the DecrementTTL option.

The option was not actually working, as it could be seen on traceroute or mtr.

The problem is that it was checking if the TTL was < 1 (so equal to 0) before decrementing it.

This meant that a packet with a TTL of 1 was being sent with a TTL of 0 on the VPN, instead of being discarded with the ICMP error message.

Cheers,
Vittorio G

### DIFF
--- a/src/route.c
+++ b/src/route.c
@@ -891,7 +891,7 @@ static bool do_decrement_ttl(node_t *source, vpn_packet_t *packet) {
 			if(!checklength(source, packet, ethlen + ip_size))
 				return false;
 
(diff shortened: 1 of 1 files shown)


VittGam:
By the way, IPv6 in switch mode does not work _by design_ when this option is enabled, because the Tinc host will then act as a pseudo-router; and the IPv6 discovery protocol has been engineered to explicitly break in presence of routers, to avoid leaks between routed subnets.

PR #97, opened 2015-09-04
Fix source IP address for ICMP unreachable packets generated by tinc.

VittGam:
Hello,

This patch fixes source IP address for ICMP unreachable packets generated by tinc.

It tries to send ICMP unreachable replies from an address assigned to the local machine, instead of the destination address of the original packet.

The address is found by looking up the route towards the sender of the packet that generated the error; in usual configurations, this is the tinc interface.

This is pretty much the same routing behaviour of the Linux kernel when sending out ICMP errors: it sets the source address according to the route to the destination, without thinking about particular source routing that might have been in effect for a normal reply. In short, it is the same of the output of `ip route get to <ip>` without `from` and `iif` arguments.

This also fixes the traceroute display in mtr when using the DecrementTTL option.

By the way, I've tested this patch only on Linux; in theory it shouldn't break any other platform that conforms to the POSIX standards, but who knows...

Cheers,
Vittorio G

### DIFF
--- a/src/route.c
+++ b/src/route.c
@@ -258,7 +258,6 @@ static void route_ipv4_unreachable(node_t *source, vpn_packet_t *packet, length_
 	struct in_addr ip_src;
 	struct in_addr ip_dst;
 	uint32_t oldlen;
-	int sockfd;
 
 	if(ratelimit(3))
 		return;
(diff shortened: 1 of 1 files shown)


VittGam:
Il 04.09.2015 16:33 Guus Sliepen ha scritto:
> On Thu, Sep 03, 2015 at 07:06:15PM -0700, Vittorio G (VittGam) wrote:
> Well, it's only for the traceroute display that you'd like to see the
> correct IP address. For the other cases of ICMP packets generated by
> tinc, just using the original destination address works perfectly fine.

Yes, you're right. But in a mesh it can be useful to enable DecrementTTL
everywhere and then use a traceroute tool to see the current topology.
This makes sense especially when using router mode, not switch mode,
since by definition in a switched layer 2 network every host should be
distant only one hop from each other.

> Your solution should work everywhere, that is very nice. You are using
> port 0, which is a somewhat strange port I have just found out. The
> Internet Protocol allows port 0 in its headers just fine. However, it is
> reserved by IANA (which normally means: don't use it), and in POSIX it
> has a special meaning when you bind to port 0. In that case it chooses a
> random port number. That means you can never actually listen on port 0.
> However, connect() on Linux at least seems to happily accept destination
> port 0 and tcpdump shows it will even send packets to port 0 if you do
> TCP or write() to the UDP socket.

I knew it meant "bind to a random port" in fact, so it's strange that instead
packets get sent to port 0 on Linux... I left it to 0 because we're not
really going to make a real UDP outgoing connection, since nothing is sent
and the socket is immediately closed.

> Apparently, many routers and firewalls treated port 0 special as well,
> sometimes being unable to block port 0. This was exploited in the past,
> so sending anything to port 0 might now be seen as an attack. Of course,
> your solution never causes a packet to be sent.

Interesting. Yeah, no packet is sent, and no connection should be tracked
in conntrack or similar subsystems.

> I don't know, some OSes might not allow connections to port 0, but in
> that case the fallback will be used, so all is well.

Yes, right.

> The only case where it might be a problem is when you run tinc on a node
> that is just a forwarder, and doesn't have a tun interface (with
> DeviceType = dummy for example). In that case you might be leaking the
> real IP address of that node.

Uhm, but then, does it make sense to enable DecrementTTL if you want that
node to be transparent in the network?

>
> --
> Met vriendelijke groet / with kind regards,
> Guus Sliepen

Cheers,
Vittorio

VittGam:
> Uhm, but then, does it make sense to enable DecrementTTL if you want that
> node to be transparent in the network?

Of course I'm wrong here, the ICMP packets can get sent even in other
conditions. Maybe a new option is the solution? Say, UseNodeAddressInICMPErrors
or something like that.

gsliepen:
On Fri, Sep 04, 2015 at 07:43:00AM -0700, Vittorio G (VittGam) wrote:

> > The only case where it might be a problem is when you run tinc on a node
> > that is just a forwarder, and doesn't have a tun interface (with
> > DeviceType = dummy for example). In that case you might be leaking the
> > real IP address of that node.
> 
> Uhm, but then, does it make sense to enable DecrementTTL if you want that
> node to be transparent in the network?

It depends on what you want to use it for; another use is to prevent
routing loops from doing too much damage.

Anyway, I like the patch but I think it would be a bit safer if it only
tries the address lookup when type == ICMP6?_TIME_EXCEEDED. Do you want
to make that change or shall I?

-- 

PR #98, opened 2015-09-06
Fix DecrementTTL option for packets destined to the local node.

VittGam:
Hello,

When there are packets destined to the local node, the current code is doing the wrong thing:

```
traceroute to host-running-tinc (192.168.3.1), 30 hops max, 60 byte packets
 1  OpenWrt.lan  2.549 ms  2.479 ms  2.420 ms
 2  192.168.3.1  70.187 ms  86.899 ms  86.933 ms
 3  192.168.3.1  87.121 ms  88.137 ms  89.909 ms
```

With this patch this is fixed:

```
traceroute to host-running-tinc (192.168.3.1), 30 hops max, 60 byte packets
 1  OpenWrt.lan  2.506 ms  2.435 ms  2.379 ms
 2  192.168.3.1  80.047 ms  89.361 ms  90.020 ms
```

And tracerouting a host behind tinc still works as intended:

```
traceroute to host-behind-tinc (192.168.3.3), 30 hops max, 60 byte packets
 1  OpenWrt.lan  2.484 ms  2.427 ms  2.374 ms
 2  192.168.3.1  76.878 ms  77.149 ms  78.366 ms
 3  192.168.3.3  84.626 ms  85.222 ms  85.829 ms
```

Cheers,
Vittorio

### DIFF
--- a/src/route.c
+++ b/src/route.c
@@ -53,8 +53,6 @@ static const size_t icmp6_size = sizeof(struct icmp6_hdr);
 static const size_t ns_size = sizeof(struct nd_neighbor_solicit);
 static const size_t opt_size = sizeof(struct nd_opt_hdr);
 
-static bool do_decrement_ttl(node_t *source, vpn_packet_t *packet);
-
 #ifndef MAX
 #define MAX(a, b) ((a) > (b) ? (a) : (b))
 #endif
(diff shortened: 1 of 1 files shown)


VittGam:
I've now corrected the patch; the logic was wrong before.

I've now tested all of these cases, and tinc behavior is now right for all of them:

- host -> `tinc1`
- host -> `tinc1` (alternate IP of tinc interface)
- host -> `tinc1` (IP of non-tinc interface, eg. `eth0`)
- host -> `tinc1` -> host in subnet of `tinc1`
- host -> `tinc1` -> `tinc2`
- host -> `tinc1` -> `tinc2` (alternate IP of tinc interface)
- host -> `tinc1` -> `tinc2` (IP of non-tinc interface, eg. `eth0`)
- host -> `tinc1` -> `tinc2` -> host in subnet of `tinc2`

gsliepen:
On Thu, Sep 24, 2015 at 08:01:50PM -0700, Vittorio G (VittGam) wrote:

> I've now corrected the patch; the logic was wrong before.
> 
> I've now tested all of these cases, and tinc behavior is now right for all of them:
> 
> - host -> `tinc1`
> - host -> `tinc1` (alternate IP of tinc interface)
> - host -> `tinc1` (IP of non-tinc interface, eg. `eth0`)
> - host -> `tinc1` -> host in subnet of `tinc1`
> - host -> `tinc1` -> `tinc2`
> - host -> `tinc1` -> `tinc2` (alternate IP of tinc interface)
> - host -> `tinc1` -> `tinc2` (IP of non-tinc interface, eg. `eth0`)
> - host -> `tinc1` -> `tinc2` -> host in subnet of `tinc2`

Great. Just two minor nitpickings: can you move the do_decrement_ttl()
function up so you don't need the forward declaration at the top of
route.c anymore, and rename broadcast_packet_helper() to
route_broadcast()?

-- 

VittGam:
For the second, no problem!

For the first however, there are at least three other functions that would need to be moved then. I'm going to see if that doesn't create circular dependency loops.

VittGam:
Hmmm, I'd have to reorder other functions so that the logical order `generic - ipv4 - ipv6 - etc` is broken. Would this be okay? In my opinion the forward declaration is cleaner in this case.

gsliepen:
On Fri, Sep 25, 2015 at 06:39:23AM -0700, Vittorio G (VittGam) wrote:

> Hmmm, I'd have to reorder other functions so that the logical order `generic - ipv4 - ipv6 - etc` is broken. Would this be okay? In my opinion the forward declaration is cleaner in this case.

I'm OK with it. I don't like forward declarations because they can cause
problems if they are not correct.

-- 

VittGam:
Done! It's building without warnings, and it still works, so it should be OK ;)

PR #100, opened 2015-11-01
Handle collisions gracefully in hash tables

dechamps:
This rewrites the hash table implementation to make the hash table an actual container that will not randomly overwrite items when collisions occur. This is done in the usual way: by making each bucket a linked list (as opposed to just a single item).

I'm doing this for two reasons. The first reason is obvious - it's to make sure we don't suddenly degrade to O(n) performance when two "hot" items end up in the same bucket.

The second, and most important, reason, has to do with `node_udp_cache`. It turns out that there is one scenario in which the contents of `node_udp_cache` has *correctness* implications, not just performance implications. This has to do with the way `handle_incoming_vpn_data()` is implemented.

Assume the following topology:

    A <-> B <-> C

Now let's consider the perspective of tincd running on B, and let's assume the following is true:

- All nodes are using the 1.1 protocol with node IDs and relaying    support.
- Nodes A and C have UDP addresses that hash to the same value.
- Node C "wins" in the `node_udp_cache` (i.e. it overwrites A in the    cache).
- Node A has a "dynamic" UDP address (i.e. an UDP address that has been detected dynamically and cannot be deduced from edge addresses).

Then, before this commit, A would be unable to relay packets through B.

This is because `handle_incoming_vpn_data()` will fall back to `try_harder()`, which won't be able to match any edge addresses, doesn't check the dynamic UDP addresses, and won't be able to match any keys because this is a relayed packet which is encrypted with C's key, not B's. As a result, tinc will fail to match the source of the packet and will drop the packet with a "Received UDP packet from unknown source" message.

I have seen this happen in the wild; it is actually quite likely to occur when there are more than a handful of nodes because `node_udp_cache` only has 256 buckets, making collisions quite likely. This problem is quite severe because it can completely prevent all packet communication between nodes - indeed, if node A tries to initiate some communication with C, it will use relaying at first, until C responds and helps A establish direct communication with it (e.g. hole punching). If relaying is broken, C will not help establish direct communication, and as a result no packets can make it through at all.

The bug can be reproduced fairly easily by reproducing the topology above while changing the (hardcoded) `node_udp_cache` size to 1 to force a collision. One will quickly observe various issues when trying to make A talk to C. Setting `IndirectData` on B will make the issue even more severe and prevent all communication.

Arguably, another way to fix this problem is to make `try_harder()` compare the packet's source address to each node's dynamic UDP addresses. However, I do not like this solution because if two "hot" nodes are contending on the same hash bucket, `try_harder()` will be called very often and packet routing performance will degrade closer to O(N) (where N is the total number of nodes in the graph). Improving the hash table data structure fixes the bug without introducing this performance problem.

### DIFF
--- a/src/hash.c
+++ b/src/hash.c
@@ -55,60 +55,76 @@ hash_t *hash_alloc(size_t n, size_t size) {
 	hash_t *hash = xzalloc(sizeof *hash);
 	hash->n = n;
 	hash->size = size;
-	hash->keys = xzalloc(hash->n * hash->size);
-	hash->values = xzalloc(hash->n * sizeof *hash->values);
+	hash->buckets = xzalloc(hash->n * sizeof *hash->buckets);
 	return hash;
 }
 
 void hash_free(hash_t *hash) {
-	free(hash->keys);
-	free(hash->values);
+	hash_clear(hash);
 	free(hash);
 }
 
 /* Searching and inserting */
 
+static hash_node_t **hash_bucket(const hash_t *hash, const void *key) {
+	return &hash->buckets[modulo(hash_function(key, hash->size), hash->n)];
+}
+
+static hash_node_t **hash_locate(const hash_t *hash, hash_node_t **node, const void *key) {
+	while (*node && memcmp((*node)->key, key, hash->size))
+		node = &(*node)->next;
+	return node;
+}
+
+static void hash_add(const hash_t *hash, hash_node_t **bucket, const void *key, const void *value) {
+	hash_node_t *node = xzalloc(sizeof *node);
+	node->key = xmalloc(hash->size);
+	memcpy(node->key, key, hash->size);
+	node->value = value;
+
+	node->next = *bucket;
+	*bucket = node;
+}
+
 void hash_insert(hash_t *hash, const void *key, const void *value) {
-	uint32_t i = modulo(hash_function(key, hash->size), hash->n);
-	memcpy(hash->keys + i * hash->size, key, hash->size);
-	hash->values[i] = value;
+	hash_add(hash, hash_bucket(hash, key), key, value);
 }
 
 void *hash_search(const hash_t *hash, const void *key) {
-	uint32_t i = modulo(hash_function(key, hash->size), hash->n);
-	if(hash->values[i] && !memcmp(key, hash->keys + i * hash->size, hash->size)) {
-		return (void *)hash->values[i];
-	}
-	return NULL;
+	hash_node_t *node = *hash_locate(hash, hash_bucket(hash, key), key);
+	return node ? ((void*)node->value) : NULL;
 }
 
 void *hash_search_or_insert(hash_t *hash, const void *key, const void *value) {
-	uint32_t i = modulo(hash_function(key, hash->size), hash->n);
-	if(hash->values[i] && !memcmp(key, hash->keys + i * hash->size, hash->size))
-		return (void *)hash->values[i];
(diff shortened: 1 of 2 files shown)


gsliepen:
On Sun, Nov 01, 2015 at 06:13:52AM -0800, Etienne Dechamps wrote:

> This rewrites the hash table implementation to make the hash table an actual container that will not randomly overwrite items when collisions occur. This is done in the usual way: by making each bucket a linked list (as opposed to just a single item).
> 
> I'm doing this for two reasons. The first reason is obvious - it's to make sure we don't suddenly degrade to O(n) performance when two "hot" items end up in the same bucket.
> 
> The second, and most important, reason, has to do with `node_udp_cache`. It turns out that there is one scenario in which the contents of `node_udp_cache` has *correctness* implications, not just performance implications. This has to do with the way `handle_incoming_vpn_data()` is implemented.
[...]

Very good. Some comments though:

Tinc uses the hash tables only for caches. The problem with adding
proper buckets is that there is no replacement of nodes anymore. You
removed hash_resize(), but actually this function should be kept and
actually made use of when the number of entries in the hash is larger
than the number of buckets. Or, we should add a kind of timestamp to the
hash entries so that you can remove a fraction of the least recently
used entries when that situation happens.

I think the naming of the functions in your commit could be a bit
better. We have hash_search(), and now you add hash_locate(), it might
be a bit confusing later on when we forgot what this code is meant to
do. Maybe hash_search_in_bucket() would be better? Similar for
hash_insert() and hash_add(), maybe rename the latter to
hash_insert_into_bucket()?

-- 

dechamps:
> Tinc uses the hash tables only for caches. The problem with adding proper buckets is that there is no replacement of nodes anymore.

Good catch. I didn't realize that hash tables were also used for the subnet cache, and there is a distinct key for every destination address of every packet that flows through the system, which means that this can blow up quite badly. So my solution is clearly not acceptable.

Maybe we should keep the old hash table implementation for the subnet cache, and upgrade `node_udp_cache` to a `splay_tree_t` (i.e. it won't be a cache anymore)? The number of addresses is bounded by the number of nodes anyway, so the number of entries can't "explode".

(Sidebar: the way this is done for node IDs is to use both a tree and a cache. I think that's overkill IMHO, and would like to write a commit to remove node_id_cache. It's more trouble that it's worth - with 1000 nodes log2(1000) is 10, so tree lookups are already plenty fast, and will be dwarfed by decryption cost anyway.)

gsliepen:
On Sun, Nov 01, 2015 at 08:07:05AM -0800, Etienne Dechamps wrote:

> > Tinc uses the hash tables only for caches. The problem with adding proper buckets is that there is no replacement of nodes anymore.
> 
> Good catch. I didn't realize that hash tables were also used for the subnet cache, and there is a distinct key for every destination address of every packet that flows through the system, which means that this can blow up quite badly. So my solution is clearly not acceptable.
> 
> Maybe we should keep the old hash table implementation for the subnet cache, and upgrade `node_udp_cache` to a `splay_tree_t` (i.e. it won't be a cache anymore)? The number of addresses is bounded by the number of nodes anyway, so the number of entries can't "explode".

Sure, just revert commit 90f1cba. Still, for Subnets you can also have
the problem that packets have a source and destination address that hash
to the same position, in which case you get O(n) behaviour again. So
buckets might still be useful, but perhaps implementing open addressing
might be better.

-- 

dechamps:
This is superseded by #101 which was merged.

PR #103, opened 2015-11-15
Add UPnP support to tinc

dechamps:
This series of commits makes tinc capable of requesting TCP and UDP port redirection from UPnP-IGD devices on the local network. This will further improve tinc's reliability and performance when nodes are behind NATs, especially home routers.

This is accomplished by introducing an optional dependency on a lightweight UPnP client library: libminiupnpc.


gsliepen:
On Sun, Nov 15, 2015 at 09:00:06AM -0800, Etienne Dechamps wrote:

> This series of commits makes tinc capable of requesting TCP and UDP port redirection from UPnP-IGD devices on the local network. This will further improve tinc's reliability and performance when nodes are behind NATs, especially home routers.
> 
> This is accomplished by introducing an optional dependency on a lightweight UPnP client library: libminiupnpc.

Very nice! Some comments:

- Is it necessary to have both "UPnPMapUDPPort" and "UPnPTCPPort"? I'd
  simplify it and just have a single "UPnP" option.

- It might be an idea to have autoconf automatically enable or disable
  support for UPnP if you don't specify --disable-upnp or --enable-upnp
  manually. For compatibility reasons, LZO and zlib support should not
  automatically be disabled, but actually, curses and readline support
  should be automatically detected as well (I was just lazy and
  copy&pasted some m4 code just like you did ;)).

- Nitpicking: in upnp_add_mapping(), maybe use identname instead of
  creating "tinc (netname)"?

- Out of curiosity, what happens if you have multiple computers, each
  running tinc on the same port and with the same netname, behind the
  same UPnP-enabled router?

-- 

dechamps:
> Is it necessary to have both "UPnPMapUDPPort" and "UPnPTCPPort"? I'd simplify it and just have a single "UPnP" option.

Well, I don't know. Personally I don't need my tinc "clients" (the ones behind NATs) to be able to accept incoming metaconnections; in fact I would prefer if the port stayed closed because it's slightly more secure that way. However, I do very much want them to have open UDP ports :) 

> It might be an idea to have autoconf automatically enable or disable support for UPnP if you don't specify --disable-upnp or --enable-upnp manually.

I disagree. I really don't like configure scripts that enable or disable features "automagically", because it means that if some library is broken or something, the next build will *silently* have missing features, which can get very annoying if you don't notice it immediately. It's better to have it fail early and hard, so that the user *has* to manually and *knowingly* disable the feature. That way it's much easier to tell what's going on and you don't have features suddenly missing from binaries for silly reasons.

> Nitpicking: in upnp_add_mapping(), maybe use identname instead of creating "tinc (netname)"?

Good catch. Done.

> Out of curiosity, what happens if you have multiple computers, each running tinc on the same port and with the same netname, behind the same UPnP-enabled router?

Good question. I think there are two possible outcomes:
- The two nodes will constantly override each other's mapping in an ugly back-and-forth, likely causing intermittent tinc connectivity loss;
- Or the UPnP device will honor the first lease and will always return an error to the second node because the first node got there first and keeps renewing the lease before it expires.

Clearly we want the latter; that would be the "least worse" outcome. However, I have no idea what's actually going to happen. Even if we looked at the UPnP spec it would not be of much help: UPnP devices are very bad at following the spec, especially for this sort of edge case (error handling in these things is absolutely terrible). Take a look at http://www.upnp-hacks.org/ for a depressing read.

To be honest, in my opinion the situation you describe is doomed to cause issues anyway, even without UPnP; that's why I introduced the "Port = 0" (dynamically assigned port) feature so that I never have to deal with this problem. All the "client" nodes on my network (i.e. the ones behind NATs) use dynamically assigned UDP ports for that reason.

gsliepen:
On Sun, Nov 15, 2015 at 02:39:32PM -0800, Etienne Dechamps wrote:

> > Is it necessary to have both "UPnPMapUDPPort" and "UPnPMapTCPPort"? I'd simplify it and just have a single "UPnP" option.
> 
> Well, I don't know. Personally I don't need my tinc "clients" (the ones behind NATs) to be able to accept incoming metaconnections; in fact I would prefer if the port stayed closed because it's slightly more secure that way. However, I do very much want them to have open UDP ports :) 

Ok. Well, I don't see how one would ever want to have TCP but not UDP
open. How about having the variable "UPnP" with the possible values
"no", "yes" and "udponly"?

> > It might be an idea to have autoconf automatically enable or disable support for UPnP if you don't specify --disable-upnp or --enable-upnp manually.
> 
> I disagree. I really don't like configure scripts that enable or disable features "automagically", because it means that if some library is broken or something, the next build will *silently* have missing features, which can get very annoying if you don't notice it immediately. It's better to have it fail early and hard, so that the user *has* to manually and *knowingly* disable the feature. That way it's much easier to tell what's going on and you don't have features suddenly missing from binaries for silly reasons.

Ok. But I'd rather not annoy people with yet another *seemingly*
mandatory dependency. Linux distributions are not the problem, but
rather users who compile tinc themselves, a lot of them savvy enough to
get the sources and run ./configure && make, but who have a hard time
solving errors during the build. Apparently it is easier to ask me than
to do ./configure --help or to read the manual...

> > Out of curiosity, what happens if you have multiple computers, each running tinc on the same port and with the same netname, behind the same UPnP-enabled router?
> 
> Good question. I think there are two possible outcomes:
> - The two nodes will constantly override each other's mapping in an ugly back-and-forth, likely causing intermittent tinc connectivity loss;
> - Or the UPnP device will honor the first lease and will always return an error to the second node because the first node got there first and keeps renewing the lease before it expires.

Do you have a UPnP IGD yourself? It would be interesting to try it out.
I don't though... but I'll see if I can try out linux-igd or something
similar.

> Clearly we want the latter; that would be the "least worse" outcome. However, I have no idea what's actually going to happen. Even if we looked at the UPnP spec it would not be of much help: UPnP devices are very bad at following the spec, especially for this sort of edge case (error handling in these things is absolutely terrible). Take a look at http://www.upnp-hacks.org/ for a depressing read.

Ugh. The fact that it's XML over HTTP over UDP is already horrible. If I
have some time I'll review the miniupnpc code to see if that might be
vulnerable to malformed responses from the IGD.

-- 

dechamps:
> Ok. Well, I don't see how one would ever want to have TCP but not UDP open. How about having the variable "UPnP" with the possible values "no", "yes" and "udponly"?

Sure. I'll make the change as soon as I have some time.

> Ok. But I'd rather not annoy people with yet another *seemingly* mandatory dependency. Linux distributions are not the problem, but rather users who compile tinc themselves, a lot of them savvy enough to get the sources and run ./configure && make, but who have a hard time solving errors during the build. Apparently it is easier to ask me than to do ./configure --help or to read the manual...

Well, I would be fine with disabling miniupnpc by default (i.e. requiring the use of `--enable-miniupnpc`). This way a simple `./configure` will work just fine and you won't get any support requests (well, only from the few people who actually want UPnP but can't figure out how to make it work).

> Do you have a UPnP IGD yourself? It would be interesting to try it out. I don't though... but I'll see if I can try out linux-igd or something similar.

I use miniupnpd. I can try running a few experiments. The thing is, I would expect UPnP implementations in random off-the-shelf routers to exhibit significantly worse behavior than publicly available, open source UPnP implementations like miniupnpd or linux-igd. (Thankfully, some of these off-the-shelf devices do use these open source implementations, but most of them don't.)

> Ugh. The fact that it's XML over HTTP over UDP is already horrible. If I have some time I'll review the miniupnpc code to see if that might be vulnerable to malformed responses from the IGD.

Mmm. That's a good point, I did not realize that my code could allow the miniupnpc library to be used as an attack vector (especially if the UPnP device is compromised). I guess I made the right call by disabling the feature by default :)

gsliepen:
On Sun, Nov 15, 2015 at 03:39:45PM -0800, Etienne Dechamps wrote:

> > Ok. Well, I don't see how one would ever want to have TCP but not UDP open. How about having the variable "UPnP" with the possible values "no", "yes" and "udponly"?
> 
> Sure. I'll make the change as soon as I have some time.
> 
> > Ok. But I'd rather not annoy people with yet another *seemingly* mandatory dependency. Linux distributions are not the problem, but rather users who compile tinc themselves, a lot of them savvy enough to get the sources and run ./configure && make, but who have a hard time solving errors during the build. Apparently it is easier to ask me than to do ./configure --help or to read the manual...
> 
> Well, I would be fine with disabling miniupnpc by default (i.e. requiring the use of `--enable-miniupnpc`). This way a simple `./configure` will work just fine and you won't get any support requests (well, only from the few people who actually want UPnP but can't figure out how to make it work).

Great.

> > Ugh. The fact that it's XML over HTTP over UDP is already horrible. If I have some time I'll review the miniupnpc code to see if that might be vulnerable to malformed responses from the IGD.
> 
> Mmm. That's a good point, I did not realize that my code could allow the miniupnpc library to be used as an attack vector (especially if the UPnP device is compromised). I guess I made the right call by disabling the feature by default :) 

Yes :) Hm, it might be even better if it really was a separate program.
Not just a fork(), but a real exec() as well so you don't have private
key material available to miniupnpc. But that's maybe something for
later.

-- 

dechamps:
Changed:

- Disabled the miniupnc dependency by default. It has to be enabled manually using `./configure --enable-miniupnpc`
- Changed the configuration to use one `UPnP` option that takes `yes`, `no` or `udponly` values.
- Improved the documentation for the `UPnP` option to make the user aware that miniupnpc can constitute a security risk.

The changes were made in the existing commits.

dechamps:
By the way, I don't think I'll be implementing NAT-PMP support. I think it's simply not worth it - as far as I can tell, the only devices that support NAT-PMP but not UPnP are Apple devices, which are seldom used as home routers (except maybe for the Airport products, but I'm guessing it probably affects a very small number of users at this point).

Adding new code that needs to be maintained, as well as an extra dependency on another network-facing library (libnatpmp) with security implications, is simply not justified for such a small number of cases IMHO.

PR #111, opened 2016-04-27
Proposal for fixing long wait time when autoconnect=yes and only few nodes have defined address

thorkill:
When AutoConnect is enabled tinc tries to connect to other nodes picking them at random.
This may be sane default behavior but it may take ages if only few nodes have
defined Address in thier config.

Proposed solution to this problem:
- Filter out nodes without known address in periodic_handler
  I have added new node->status.has_known_address bool


gsliepen:
I implemented your approach in the 1.1 branch.

PR #116, opened 2016-05-09
Allow PKT_PROBE in sptps_send_record

thorkill:
sptps_send_record prevented PKT_PROBE to be send in send_sptps_packet.
This occurred mostly when data was on "the wire" for some subnet.
route() would then trigger try_tx/try_udp which would be dropped by
sptps_send_record producing annoying amount of "Handshake phase
not finished yet" log messages.

### DIFF
--- a/src/sptps.c
+++ b/src/sptps.c
@@ -128,11 +128,13 @@ static bool send_record_priv(sptps_t *s, uint8_t type, const void *data, uint16_
 
 // Send an application record.
 bool sptps_send_record(sptps_t *s, uint8_t type, const void *data, uint16_t len) {
+
+	if((!s->outstate) && (type != PKT_PROBE))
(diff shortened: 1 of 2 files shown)


src/sptps.c
```diff
@@ -128,11 +128,13 @@ static bool send_record_priv(sptps_t *s, uint8_t type, const void *data, uint16_
 
 // Send an application record.
 bool sptps_send_record(sptps_t *s, uint8_t type, const void *data, uint16_t len) {
+
+	if((!s->outstate) && (type != PKT_PROBE))
+		return error(s, EINVAL, "Handshake phase not finished yet (outstate:%x type: %d)",
```
thorkill: You can of course remove the outstate, type information in this commit - tinc-ponyhof does have excesive debugging informations.

gsliepen:
Cherry-picked rather than merged :)

PR #117, opened 2016-05-10
Prevent receive_tcppacket_sptps from forwarding data when in handshake

thorkill:
Prevent receive_tcppacket_sptps from forwarding data to another nodes when we are in handshake phase with those nodes

Forwarding data to node which does not have vaildkey yet would destroy the running handshake.


splitice:
@gsliepen I think this is worth a review. Easy enough to make merge.

It looks sane enough.

gsliepen:
Fixed by commit 2c9126ab604fce80d9acd1bbdb485eae50aefcad.

PR #126, opened 2016-10-12
[1.1] route: Support ToS/DiffServ priority inheritance when routing IPv6 packets.

VittGam:
This is the patch for tinc 1.1.

Cheers,
Vittorio


gsliepen:
Merged, thanks!

Issue #133, opened 2016-11-12
simplify systemd service file


fkooman:
After running into some issues with the tinc systemd integration where I was unable to make it wait for the network to be online (on CentOS 7), see e.g. https://bugzilla.redhat.com/show_bug.cgi?id=1394512 I propose to remove `tinc.service` and modify `tinc@.service` to the following, this is modelled after the `openvpn@.service` file:

```
[Unit]
Description=Tinc net %i
After=network.target

[Service]
Type=simple
WorkingDirectory=/etc/tinc/%i
ExecStart=/usr/sbin/tincd -n %i -D
ExecReload=/usr/sbin/tincd -n %i -kHUP
TimeoutStopSec=5
Restart=always
RestartSec=60

[Install]
WantedBy=multi-user.target
```

aflyhorse:
since journald automatically separate logs from each service, I recommend adding '-d 1' to show more details regarding connection failed, re-established, etc.

fkooman:
It could also be 0 or 2, depending... 2 can also be very useful, at one point I forgot to chmod +x tinc-up and it didn't work...

```
DEBUG LEVELS
     The tinc daemon can send a lot of messages to the syslog.  The higher the debug level, the more mes‐
     sages it will log.  Each level inherits all messages of the previous level:

     0       This will log a message indicating tincd has started along with a version number.  It will
             also log any serious error.

     1       This will log all connections that are made with other tinc daemons.

     2       This will log status and error messages from scripts and other tinc daemons.

     3       This will log all requests that are exchanged with other tinc daemons. These include authenti‐
             cation, key exchange and connection list updates.

     4       This will log a copy of everything received on the meta socket.

     5       This will log all network traffic over the virtual private network.
```

Hoeze:
Hi,
I've got another safety related improvement:
`ExecStart=/usr/sbin/tincd -n %i -D **-R -U nobody**`
This would automatically chroot tincd to the config directory and drop privileges to 'nobody' user.

Works for me on Debian Stretch + ArchLinux.

gsliepen:
That might work for you, but it might not work for everybody. In particular: the chroot environment has to be prepared so hostname resolution continues to work and files must be readable by user nobody. But also if one has more scripts than just tinc-up (which is the only one executed before the setuid() call), then user nobody is probably not what you want. So unfortunately I cannot add that without breaking a lot of people's setup.

tuxcrafter:
[...]

PR #134, opened 2016-11-17
simplify systemd unit file

fkooman:
PR for issue #133.


aflyhorse:
IMHO, the original "tinc.service" seems to act as a general all-in-one switch for all other tinc services.
I'm not sure if removing is proper.

fkooman:
Ah, if that is a feature that is being used then that makes sense... It seems it is based on the Debian OpenVPN systemd unit. 

Do you have any idea how to fix the issue on CentOS where the current service file does not wait for the network to be up to start? https://bugzilla.redhat.com/show_bug.cgi?id=1394512.

aflyhorse:
I thought "After=network.target" should work. On the other side, tweaks must be made in actual C/C++ code, to make it more friendly to network freak.
```
If you are a developer, instead of wondering what to do about network.target, please just fix your program to be friendly to dynamically changing network configuration. That way you will make your users happy because things just start to work, and you will get fewer bug reports as your stuff is just rock solid.
```
source: https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/

heroxbd:
Shall close this PR?

BTW, greetings @aflyhorse !

aflyhorse:
I'm suffering from that tinc@xxx.service doesn't start automatically in CentOS 7.x. It's working nice in Fedora 25+ however, so it might be another problem. Seems the distro files have be rewritten, @fkooman are you still interested in the PR?

BTW, Greetings @heroxbd , thank you for introducing tinc to me. I've been using it to bypass numerous firewalls and routers.

tomberek:
Close?

fkooman:
I'm currentl not using tinc, so maybe it is better if someone else took over, or indeed close the PR.

aflyhorse:
The problem in CentOS 7.x has already fixed. I suggest keep the current status and close this PR.

fangfufu:
Closing the PR, as per suggestion by @tomberek and @aflyhorse , and also @fkooman no longer uses tinc.

PR #136, opened 2016-12-03
Fix bugs related to Windows nodes waking up from sleep

dechamps:
On Windows, I have noticed that tinc 1.1 sometimes has difficulties resuming operation after the computer wakes up from sleep. The issues were easy to reproduce, and I found two bugs that explain the problems.


igel-kun:
Sorry for the necro, but tinc-1.1pre17 still has problems waking properly from suspend.

PR #141, opened 2017-03-01
Add fd_device

Pacien:
For Android stuff.
Tested with Android 5.1.1 VPN API.


gsliepen:
I think you missed something in your commit; I get some errors trying to compile it:

    fd_device.c: In function ‘get_ip_ethertype’:
    fd_device.c:74:10: error: ‘ETHERTYPE_MAX’ undeclared (first use in this function)
       return ETHERTYPE_MAX;

Also, it might be nice if you could rebase your work onto the latest commit on the 1.1 branch.

Pacien:
Thank you for your feedback.

`ETHERTYPE_MAX` is apparently defined only on Android and some BSD.
I replaced it with `ETH_P_MAX` internally defined.

Also just rebased on the latest commit.

PR #154, opened 2017-10-03
Allow compilation from a build directory

nemunaire:
Currently, it is not possible to build the project from a separate directory than the one containing sources.

In fact:
```
$ make distcheck
[...]
  CC       avl_tree.o
In file included from ../../src/avl_tree.c:33:0:
../../src/system.h:24:23: fatal error: ../config.h: No such file or directory
 #include "../config.h"
```

This patch fixes that and then avoid creating exceptional behavior for `tinc` in some automated packager that relies on clean separate builds.

Moreover, one last error was due to the use of wildcard in `m4/Makefile.am` whereas [it is not supported by design](https://www.gnu.org/software/automake/manual/html_node/Wildcards.html).

Thanks for your project!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3,7 +3,7 @@ dnl Process this file with autoconf to produce a configure script.
 AC_PREREQ(2.61)
 AC_INIT([tinc], [1.0.32])
 AC_CONFIG_SRCDIR([src/tincd.c])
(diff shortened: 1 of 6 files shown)


dechamps:
Related: #77 (but mine was for the 1.1 branch)

gsliepen:
Merged, thanks! I've also looked at #77, and copied the way it's doing #include $srcdir/src/have.h in autoconf.ac, instead of saving/restoring the CPPFLAGS.

PR #166, opened 2018-01-04
Support MSS clamping for IP in IP (RFC 2003) packets

dechamps:
This change allows tinc MSS clamping to operate on TCP streams that are inside an IP in IP tunnel.


gsliepen:
It's missing a check that `packet->len > start + 20` after you adjust start, otherwise subsequent if statements might right after the end of the packet.

dechamps:
Right. Fixed.

PR #174, opened 2018-02-01
Reset the write event before each call to WriteFile()

millert:
In device_issue_read() there is no need to reset Offset and OffsetHigh to 0; they are only used for seekable files (not sockets).

Reset the write event before the call to WriteFile().  This is consistent with how the read event is reset before ReadFile().

Clear device_write_packet.len() if WriteFile() fails with an error other than ERROR_IO_PENDING; otherwise write_packet() will call GetOverlappedResult() the next time it is run even though there is no write in progress.

### DIFF
--- a/src/mingw/device.c
+++ b/src/mingw/device.c
@@ -48,9 +48,6 @@ static const char *device_info = "Windows tap device";
 extern char *myport;
 
 static void device_issue_read() {
-	device_read_overlapped.Offset = 0;
-	device_read_overlapped.OffsetHigh = 0;
-
 	int status;
 
 	for(;;) {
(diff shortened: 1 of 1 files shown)


dechamps:
Thank you for fixing my code :) 

I wonder if the potential bug you found on the write path for errors other than `ERROR_IO_PENDING` could explain why I've witnessed tinc sometimes getting stuck when the computer is waking up from sleep. It's difficult to confirm though, because reproducing the issue has proven difficult thus far.

dechamps:
I did some more investigation, and I can indeed confirm that this specific commit fixes an issue where, if the computer is put to sleep while there there is traffic on the tinc interface, tinc will get stuck on wakeup with the following errors in the log:

> Error while checking previous write to Windows tap device: (996) Overlapped I/O event is not in a signaled state.

I can reliably reproduce the issue by stressing the interface using iperf while putting the computer to sleep at the same time. I was unable to reproduce it again with your fix in place.

@gsliepen: you might want to mention that bugfix when writing the release notes for the next version.

PR #194, opened 2018-06-13
Expose traffic stats and UDP RTT

volth:
Expose traffic stats (rx/tx packets and bytes) and UDP round-trip-times to external programs (including Tinc CLI).

Why:
1. Monitoring programs (```collectd```, ...) could collect the information, draw graphs, sends alerts

2. In the future, the nodes advertising the same subnet with the same weight could be sorted by UDP RTT (instead of alphanumerically as now) forming poor man anycast.
For example if all nodes running DNS server would advertise ```10.8.8.8``` in the Tinc network, the requests sent to ```10.8.8.8``` would be routed to the closer node instead of the first in the alphabet.

3. Keeping track of UDP RTT would allow to detect slow paths in underlying network.
It happens that ```A->B``` is 10ms and ```B->C``` is 10ms, but ```A->C``` is 200ms.
Currently there is no fix for that at the Tinc level but at least it would make the problem visible (and allows to solve it at least by sending support tickets to the ISP)

### DIFF
--- a/src/node.c
+++ b/src/node.c
@@ -212,7 +212,7 @@ bool dump_nodes(connection_t *c) {
 		}
 
 		id[sizeof(id) - 1] = 0;
-		send_request(c, "%d %d %s %s %s %d %d %d %d %x %x %s %s %d %d %d %d %ld %ld %lu %lu %lu %lu", CONTROL, REQ_DUMP_NODES,
+		send_request(c, "%d %d %s %s %s %d %d %d %d %x %x %s %s %d %d %d %d %ld %d %"PRIu64" %"PRIu64" %"PRIu64" %"PRIu64, CONTROL, REQ_DUMP_NODES,
 		             n->name, id, n->hostname ? : "unknown port unknown",
 #ifdef DISABLE_LEGACY
 		             0, 0, 0,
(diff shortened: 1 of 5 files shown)


src/node.c
```diff
@@ -220,7 +221,7 @@ bool dump_nodes(connection_t *c) {
 #endif
 		             n->outcompression, n->options, bitfield_to_int(&n->status, sizeof(n->status)),
 		             n->nexthop ? n->nexthop->name : "-", n->via ? n->via->name ? : "-" : "-", n->distance,
-		             n->mtu, n->minmtu, n->maxmtu, (long)n->last_state_change);
+		             n->mtu, n->minmtu, n->maxmtu, (long)n->last_state_change, (long)n->udp_ping_rtt);
```
gsliepen: The cast to `long` is unnecessary, `udp_ping_rtt` is already a `long int`.

src/node.h
```diff
@@ -92,6 +92,7 @@ typedef struct node_t {
 
 	struct timeval udp_reply_sent;          /* Last time a (gratuitous) UDP probe reply was sent */
 	struct timeval udp_ping_sent;           /* Last time a UDP probe was sent */
+	long int udp_ping_rtt;                  /* Round trip time of UDP ping (in microseconds) */
```
gsliepen: Actually, this can just be a regular `int`. 32 bits is already enough to handle RTTs of 35 minutes, by which time we've already sent another probe or have given up on the node altogether.

src/tincctl.c
```diff
@@ -1331,8 +1332,11 @@ static int cmd_dump(int argc, char *argv[]) {
 					continue;
 				}
 
-				printf("%s id %s at %s port %s cipher %d digest %d maclength %d compression %d options %x status %04x nexthop %s via %s distance %d pmtu %d (min %d max %d)\n",
+				printf("%s id %s at %s port %s cipher %d digest %d maclength %d compression %d options %x status %04x nexthop %s via %s distance %d pmtu %d (min %d max %d)",
 				       node, id, host, port, cipher, digest, maclength, compression, options, status_int, nexthop, via, distance, pmtu, minmtu, maxmtu);
+				if (udp_ping_rtt != -1)
+					printf(" rtt %ld.%03ld", udp_ping_rtt/1000, udp_ping_rtt%1000);
```
gsliepen: Add spaces around operators. Or run `make astyle` before committing the code :)

volth:
Fix for one more compiler's format warning: https://github.com/volth/tinc/commit/4c27f8474fa8510fec7e0b2716a2a809e5660936

Issue #199, opened 2018-08-06
Multiple Address variables do not work, the first entry is always used


volth:
from [the docs](https://www.tinc-vpn.org/documentation/tinc.conf.5)
```
Address = address [port] [recommended]
             The IP address or hostname of this tinc daemon on the real network.  This will only be used when trying to make an
             outgoing connection to this tinc daemon.  Optionally, a port can be specified to use for this address.  Multiple
             Address variables can be specified, in which case each address will be tried until a working connection has been
             established.
```

It seems convenient to specify IPv4 and IPv6 addresses and fall back in case if one of the networks (4 or 6) is down.
But apparently, only the first ```Address``` variable is always used, if meta-connection to the first ```Address``` is timed out or rejected, ```tinc``` gives up and do not try the second ```Address```. Instead it always try to connect to the first address incrementing the timeout between tries. The second ```Address``` is never tried.

igel-kun:
Seconded, please consider merging the patch into the next pre-release, cheers @volth for the effort :)

gsliepen:
Fixed in 1cf2b56.

PR #200, opened 2018-08-06
address_cache: read multiple Address variables from config file

volth:
Fixes https://github.com/gsliepen/tinc/issues/199


gsliepen:
Thanks for the pull request. It's fixed in 1cf2b56f using a different approach.

igel-kun:
@volth your patch incurrs a segfault if ```str2addrinfo(address, port, SOCK_STREAM);``` returns NULL (line 172), which happens if name resolution fails

volth:
That might be, I am using IP everywhere, not domains.
Anyway, @gsliepen already fixed the original problem another way so my patch is deprecated.

PR #213, opened 2018-10-18
Set interface up before running ip route on linux

iczero:
iproute2 expects the interface to be up before adding routes, or
the error "Error: Device for nexthop is not up." happens


gsliepen:
Fixed in 5f76bc0 by using the onlink flag, which should handle even more corner-cases.

PR #215, opened 2018-11-19
Generate tinc-up.bat on windows

DarkShadow44:
On windows a file called "tinc-up" won't be executed - it needs to be named tinc-up.bat. Since it is also documented like this, we should generate a bat file when using "tinc join" on windows.


gsliepen:
This has been cherry picked as commit 2bc4752. Thanks for the patch!

PR #221, opened 2019-04-16
Fix a strict aliasing violation in inet_checksum()

maciejsszmigiero:
inet_checksum() accesses packet data as an array of uint16_t, but the
packet data can be for example of "anonymous struct pseudo" type from
route_ipv6_unreachable().
This type isn't a compatible type with uint16_t so a strict aliasing
violation occurs and causes the checksum to be computed incorrectly.

This was previously worked around by commit 7c73cb3 that thought this was
a GCC bug, however it really isn't.

Fix this by using the memcpy() idiom to read the packet data as an array of
uint16_t in inet_checksum() (this should be understood by compilers and
optimized accordingly, so no actual copy occurs).

### DIFF
--- a/src/route.c
+++ b/src/route.c
@@ -59,7 +59,6 @@ static const size_t opt_size = sizeof(struct nd_opt_hdr);
 #define MAX(a, b) ((a) > (b) ? (a) : (b))
 #endif
 
-volatile int dummy;
 static timeout_t age_subnets_timeout;
 
 /* RFC 1071 */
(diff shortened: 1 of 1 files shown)


src/route.c
```diff
@@ -64,16 +64,18 @@ static timeout_t age_subnets_timeout;
...
 static uint16_t inet_checksum(void *data, int len, uint16_t prevsum) {
-	uint16_t *p = data;
+	uint16_t word;
 	uint32_t checksum = prevsum ^ 0xFFFF;
 
 	while(len >= 2) {
-		checksum += *p++;
+		memcpy(&word, data, sizeof(word));
+		checksum += word;
+		data += 2;
```
gsliepen: This fixes the strict aliasing, but now you're doing pointer arithmetic on a `void *`, which is not legal in C or C++, and in fact just a GCC extension. So maybe declare `uint8_t *p = data` and use that?

maciejsszmigiero: You are right - I have updated the patch now to avoid using a GCC extension.

mark-stopka:
@gsliepen, the change request has been reviewed and the pull request has no conflicts with your branch.

gsliepen:
Merged into tinc 1.1. Thanks for the patch, apologies for the delay.

PR #229, opened 2019-09-24
Add export public ed25519 key command to tinc cli

admincheg:
Old OS haven't native builds of openssl 1.1, but if they have backported packages - it's a big headache at maintenance.

Tinc can be builded statically for old systems and with that additional cli command we can operate with ed25519 keys without openssl 1.1. (At this moment we can generate key pair, but we can't get public key from private)


gsliepen:
But you can already export the Ed25519 public key with this command:
```
tinc -n <netname> get Ed25519PublicKey
```
So I do not see a reason to add the command you propose. Unless there is a reason why the above command cannot be used?

admincheg:
Tinc client using a host card. If you have only private key - you can't get public without external program
2019-09-26-055006_1917x1076_scrot
.

stronny:
This command is meant to be analogous to something like `openssl rsa -in priv.pem -pubout`. We have no way to do that with command-line tools, and tinc already has all the necessary infrastructure to make it really easy. Arguably it doesn't have much to do with tinc itself, as it's just a generic crypto call in its essense, but we don't see a better place for it.

Maybe we can introduce another binary like `tinc-utils`, but that also has its overhead.

PR #242, opened 2020-04-12
Warn if system capped requested socket buffer sizes

nh2:
This helps noticing when the system does not apply the requested sizes.

### DIFF
--- a/src/net_socket.c
+++ b/src/net_socket.c
@@ -288,9 +288,39 @@ int setup_vpn_in_socket(const sockaddr_t *sa) {
 	if(udp_rcvbuf && setsockopt(nfd, SOL_SOCKET, SO_RCVBUF, (void *)&udp_rcvbuf, sizeof(udp_rcvbuf))) {
 		logger(DEBUG_ALWAYS, LOG_WARNING, "Can't set UDP SO_RCVBUF to %i: %s", udp_rcvbuf, sockstrerror(sockerrno));
 	}
+	{
+		// The system may cap the requested buffer size.
+		// Read back the value and check if it is now as requested.
+		int udp_rcvbuf_actual = -1;
+		socklen_t optlen = sizeof(udp_rcvbuf_actual);
+		if (getsockopt(nfd, SOL_SOCKET, SO_RCVBUF, (void *)&udp_rcvbuf_actual, &optlen)) {
+			logger(DEBUG_ALWAYS, LOG_WARNING, "Can't read back UDP SO_RCVBUF: %s", sockstrerror(sockerrno));
+		} else if(optlen != sizeof(udp_rcvbuf_actual)) {
+			logger(DEBUG_ALWAYS, LOG_WARNING, "Can't read back UDP SO_RCVBUF: Unexpected returned optlen %jd", (intmax_t) optlen);
+		} else {
+			if(udp_rcvbuf_actual != udp_rcvbuf) {
+				logger(DEBUG_ALWAYS, LOG_WARNING, "Can't set UDP SO_RCVBUF to %i, the system set it to %i instead", udp_rcvbuf, udp_rcvbuf_actual);
+			}
+		}
+	}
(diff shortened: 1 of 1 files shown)


nh2:
There are also some things I don't understand. When I run with this with 

```
UDPRcvBuf = 25000000
UDPSndBuf = 25000000
```

in the `tinc-1.1pre17` on config on Linux `5.4.27`, I see:

```
Can't set UDP SO_RCVBUF to 25000000, the system set it to 50000000 instead
Can't set UDP SO_SNDBUF to 25000000, the system set it to 50000000 instead
Can't set UDP SO_RCVBUF to 25000000, the system set it to 50000000 instead
Can't set UDP SO_SNDBUF to 25000000, the system set it to 50000000 instead
```

The same thing happens with the default config:

```
Can't set UDP SO_RCVBUF to 1048576, the system set it to 425984 instead
Can't set UDP SO_SNDBUF to 1048576, the system set it to 425984 instead
Can't set UDP SO_RCVBUF to 1048576, the system set it to 425984 instead
Can't set UDP SO_SNDBUF to 1048576, the system set it to 425984 instead
```

So it looks like the system set it to 2x the requested amount. Why would that be?

src/net_socket.c
```diff
@@ -288,10 +288,40 @@ int setup_vpn_in_socket(const sockaddr_t *sa) {
...
 	}
+	{
+		// The system may cap the requested buffer size.
+		// Read back the value and check if it is now as requested.
+		int udp_rcvbuf_actual = -1;
+		socklen_t optlen = sizeof(udp_rcvbuf_actual);
+		if (getsockopt(nfd, SOL_SOCKET, SO_RCVBUF, (void *)&udp_rcvbuf_actual, &optlen)) {
+			logger(DEBUG_ALWAYS, LOG_WARNING, "Can't read back UDP SO_RCVBUF: %s", sockstrerror(sockerrno));
+		} else if(optlen != sizeof(udp_rcvbuf_actual)) {
+			logger(DEBUG_ALWAYS, LOG_WARNING, "Can't read back UDP SO_RCVBUF: Unexpected returned optlen %i", optlen);
```
gsliepen, line 299: `optlen` is a `socklen_t`, you cannot assume it is the same as an `int`. So `%i` is not safe to use, unless you explicitly cast `optlen` to `int`.

gsliepen (changes requested):

src/net_socket.c
nh2: Thanks, addressed as described on https://stackoverflow.com/questions/39417560/correct-format-specifiers-for-custom-integer-types-in-c-to-ensure-portability

src/net_socket.c
```diff
@@ -288,10 +288,40 @@ int setup_vpn_in_socket(const sockaddr_t *sa) {
...
 	}
+	{
+		// The system may cap the requested buffer size.
+		// Read back the value and check if it is now as requested.
+		int udp_sndbuf_actual = -1;
+		socklen_t optlen = sizeof(udp_sndbuf_actual);
+		if (getsockopt(nfd, SOL_SOCKET, SO_SNDBUF, (void *)&udp_sndbuf_actual, &optlen)) {
+			logger(DEBUG_ALWAYS, LOG_WARNING, "Can't read back UDP SO_SNDBUF: %s", sockstrerror(sockerrno));
+		} else if(optlen != sizeof(udp_sndbuf_actual)) {
+			logger(DEBUG_ALWAYS, LOG_WARNING, "Can't read back UDP SO_SNDBUF: Unexpected returned optlen %i", optlen);
```
gsliepen, line 318: You missed the second `optlen` :)

nh2: Ah, sorry. It's fixed now.

gsliepen:
Thanks, cherry-picked as 73a45af5c50d58867efc7f1778d7c2f9906f50d9.

PR #251, opened 2020-09-08
[android] tincctl: restrict umask argument for FORTIFY

pacien:
`umask(mode)` calls that do not verify `(mode & 0777) == mode` are
rejected when the libc FORTIFY checks are enabled [1].

The unrestricted `~perms` was indeed making this assertion fail.

[1]: https://android.googlesource.com/platform/bionic/+/refs/tags/android-11.0.0_r3/libc/bionic/fortify.cpp#404

---

This is needed for tinc to correctly run when built with the latest Android NDK,
which enables the FORTIFY checks.


gsliepen:
Cherry-picked as commit 4cc4b9bcce74b3e7ec4fe539366b7cfc3b472c07

Issue #230, opened 2019-10-07
Segfault with x86_64-w64-mingw-w64


wolfy1339:
I have determined via `git bisect` that d6b45d005530496e48325a6174ecdd889a17bfc1 is the commit that introduces the segfault.
The produced binary runs for a couple seconds then it segfaults
`1.1-pre17` compiles and runs fine

wolfy1339:
`x86_64-w64-mingw32.static-gcc (GCC) 5.5.0` provided by [mxe](https://mxe.cc/)

wolfy1339:
I just compiled a newer GCC just to test and 9.2.0 is the same

arsh0r:
after [these](https://github.com/arsh0r/tinc/commit/b8bf6cc7b8da33a5d74a8eba4ab93f55a310184a) changes it compiles and runs for me.

fangfufu:
@arsh0r , thanks for providing the patch. I generated a pull request for you. This needs to be reviewed properly before merging.

fangfufu:
Fixed via https://github.com/gsliepen/tinc/pull/263#issuecomment-869172119

PR #263, opened 2021-06-23
Pull request to fix issue #230 (Segfault with x86_64-w64-mingw-w64)

fangfufu:
This pull request fix the problem described in issue https://github.com/gsliepen/tinc/issues/230.

> **Segfault with x86_64-w64-mingw-w64**
> I have determined via git bisect that d6b45d0 is the commit that introduces the segfault.
The produced binary runs for a couple seconds then it segfaults
1.1-pre17 compiles and runs fine


gsliepen:
This pull request is in a bad state: it's adding lots of files that shouldn't be there, and there were merge conflicts. I've picked the few things I think were relevant for the 1.1 branch and merged those individually.

Issue #110, opened 2016-03-26
High CPU load on cloud VMs


nh2:
This weekend I benchmarked Tinc (multiple versions) on the smallest [DigitalOcean](https://www.digitalocean.com/pricing/) instance.

I found that it generates very high CPU load (much more than on a desktop), and network performance maxes out at around 200 MBit/s with `iperf3` from one Tinc node to another, when the raw interface can deliver 1000 Mbit/s without noticeable CPU load.

This is a write up of the little investigation I did, maybe as a help if some third party Tinc contributor wants to try and solve this problem (or if I find the time for it myself at some point).

Note that I'm extremely new to Tinc (I started reading into the code this weekend), so some of this may not be 100% accurate, but @gsliepen was extremely helpful in answering my questions on the `#tinc` IRC channel - thanks for this.

There seem to be two things that make Tinc slow on these machines:

* **Encryption:**
  * The latest Tinc `1.1` branch switched from AES+SHA to ChaCha-Poly1305 (also, the `Cipher` and `Digest` config options are no longer configurable and are ignored). While ChaCha is faster than AES on standard machines, it cannot compete with the AES-NI hardware acceleration for AES that many modern CPU (and also those cloud machines) provide; the speed difference is approximately 4x (can also be tested with Tinc's `sptps_speed` utility).
  * ChaCha takes around 65% of CPU time in `perf` (make sure to use `CFLAGS=-fno-omit-frame-pointer ./configure` for measuring this).
  * Re-enabling AES would make the symmetric encryption overhead pretty much negligible, at least on Gigabit-Link machines.
  * But according to @gsliepen, this is not so easy / there is trade-off involved, because the OpenSSL interface that Tinc used in the past for AES(-NI) is not quite stable. Maybe [libsodium](https://github.com/jedisct1/libsodium/blob/89a6f58e4ddd52c8a72273d9ba9c98f1828a8c02/src/libsodium/crypto_aead/aes256gcm/aesni/aead_aes256gcm_aesni.c) could be used instead? [No software fallback](https://github.com/jedisct1/libsodium/tree/89a6f58e4ddd52c8a72273d9ba9c98f1828a8c02/src/libsodium/crypto_aead/aes256gcm) seems to be provided though.
* **Syscalls:**
  * When watching Tinc in `htop` while `iperf3` is running, there's a lot of red, meaning time spent in the kernel. This is even more prominent when `Cipher = none` and `Digest = none`, which is still possible in Tinc 1.0.
[...]

PR #268, opened 2021-06-29
bail out of logging early

splitice:
**Very** Low hanging fruit. 

Accounts for 2% of RX perf trace on Vultr Single Core 1GB VM.

### DIFF
--- a/src/logger.c
+++ b/src/logger.c
@@ -37,23 +37,22 @@ static FILE *logfile = NULL;
 static HANDLE loghandle = NULL;
 #endif
 static const char *logident = NULL;
-bool logcontrol = false;
+bool logcontrol = false; // controlled by REQ_LOG <level>
 int umbilical = 0;
+
+static bool should_log(int level) {
+	return (level <= debug_level && logmode != LOGMODE_NULL) || logcontrol;
+}
 
 static void real_logger(int level, int priority, const char *message) {
 	char timestr[32] = "";
 	static bool suppress = false;
 
-	// Bail out early if there is nothing to do.
 	if(suppress) {
 		return;
 	}
 
-	if(!logcontrol && (level > debug_level || logmode == LOGMODE_NULL)) {
-		return;
-	}
-
-	if(level <= debug_level) {
+	if (level <= debug_level) {
 		switch(logmode) {
 		case LOGMODE_STDERR:
 			fprintf(stderr, "%s\n", message);
@@ -72,17 +71,17 @@ static void real_logger(int level, int priority, const char *message) {
 			break;
 
 		case LOGMODE_SYSLOG:
-#ifdef HAVE_MINGW
+	#ifdef HAVE_MINGW
 			{
 				const char *messages[] = {message};
 				ReportEvent(loghandle, priority, 0, 0, NULL, 1, 0, messages, NULL);
 			}
 
-#else
-#ifdef HAVE_SYSLOG_H
+	#else
+	#ifdef HAVE_SYSLOG_H
 			syslog(priority, "%s", message);
-#endif
-#endif
+	#endif
+	#endif
 			break;
 
 		case LOGMODE_NULL:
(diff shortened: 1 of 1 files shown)


splitice:
Found while profiling for #110

src/logger.c
```diff
@@ -40,59 +40,62 @@ static const char *logident = NULL;
...
 int umbilical = 0;
 
+static bool should_log(int level){
+	// Bail out early if there is nothing to do.
+
+	if(!logcontrol && (level > debug_level || logmode == LOGMODE_NULL)) {
+		return false;
+	}
+
+	return (level > debug_level);
```
gsliepen, line 50: That should be `return level <= debug_level`.

src/logger.c
```diff
@@ -125,6 +128,8 @@ void logger(int level, int priority, const char *format, ...) {
 	va_list ap;
 	char message[1024] = "";
 
+	if(!should_log(level)) return;
```
gsliepen, line 131: Please run `make astyle` from the top directory.

gsliepen (changes requested):
With this patch, nothing is logged anymore on my machines.

src/logger.c
splitice: how did I miss this :(

fixed.

src/logger.c
splitice: done

src/logger.c
```diff
@@ -40,59 +40,62 @@ static const char *logident = NULL;
...
 int umbilical = 0;
 
+static bool should_log(int level) {
+	// Bail out early if there is nothing to do.
+
+	if(!logcontrol && (level > debug_level || logmode == LOGMODE_NULL)) {
+		return false;
+	}
+
+	return level <= debug_level;
```
gsliepen, line 50: This turns out to be wrong too. If `logcontrol == true`, the value of `level` is not necessarily the same debug level requested over a control channel. So we should `return true` here, or just simplify this whole function to:

```
return (level <= debug_level && logmode != LOGMODE_NULL) || logcontrol;
```

src/logger.c
```diff
@@ -40,59 +40,62 @@ static const char *logident = NULL;
...
-		return;
-	}
-
-	if(level <= debug_level) {
-		switch(logmode) {
-		case LOGMODE_STDERR:
-			fprintf(stderr, "%s\n", message);
-			fflush(stderr);
-			break;
+	switch(logmode) {
```
gsliepen, line 61: There needs to be an `if (level <= debug_level)` around this switch and the first `if`-statement after it, in order to correctly handle the case where `logcontrol == true`.

gsliepen (changes requested):

src/logger.c
splitice: Updated with the code you provided.

src/logger.c
splitice: FYI at no point in the code is `logcontrol` documented. Unless you were a seasoned developer I would expect it to be an unknown code quirk.

splitice: And is there a reason that REQ_LOG doesnt just change the log level? Slight difference in behaviour but simpler and potentially more powerful.

gsliepen: You're right that the function of `logcontrol` is not well documented. There are multiple log outputs possible at the same time, with different log levels. The normal log output (to `stdout`, a file or to the syslog) is controlled by the global variable `debug_level`. However, you can use the `tinc log <level>` command to connect to a running tinc daemon and get a stream of log messages with a different log level. Whenever there is at least one control connection requesting log information, `logcontrol` is `true`. That's when we have to do extra work for every call to `logger()` to find out which connections want which log message.

gsliepen: There's also `REQ_SET_DEBUG` which allows you to change the global log level of a running tinc daemon.

gsliepen:
Cherry-picked as a459e57de5fa4d32ee1162e49770fd8730e3707d, thanks!

Issue #252, opened 2020-09-21
send_udp_probe_packet expects dynamically sized buffer larger than the statically sized buffer, smashes stack during initialization of packet with random data


deliciouslytyped:
Moving from a specific physical wifi network to another (and removing my laptop from the dock (TODO: rule this out)) causes the tincd I have running in a container (TODO: rule this out) to segfault reproducibly, but I haven't been able to isolate the bug. It consistently gets up to the `Sending 70 bytes of metadata to [redacted] ([redacted] port 20)` portion of the attached log, and then (obviously at some point after this) segfaults.

Starting tincd with gdb results in the stack being incomprehensible (to me).
Starting it with `rr record` results in a stack smashing notification, which probably explains the previous point.
However due to an `rr replay` issue I haven't been able to yield any good info from that either.
Maybe valgrind is worth a shot? 

Here is a log:
```
[root@gate:~]# /nix/store/yairbna66xfrib2snjdq9z5gxqxlk4bw-rr-5.3.0/bin/rr record --chaos /nix/store/3218jrqqrh8cbbnhijhh2vxw2m9vm85m-tinc-1.1pre17/bin/tincd -D -U gate -n gate --pidfile /run/gate.pid -d4                                              [166/1678]
rr: Saving execution to trace directory `/root/.local/share/rr/tincd-6'.                                                                                                                                                                                                       
tincd 1.1pre17 (Jan  1 1980 00:00:00) starting, debug level 4                                                                                                                                                                                                                  
Error reading RSA private key file `/etc/tinc/gate/rsa_key.priv': No such file or directory                                             
Create an RSA keypair with `tinc -n gate generate-rsa-keys'.                                                                                                                                                                                                                    
Support for legacy protocol disabled.                                                                                                                                                                                                                                          
/dev/net/tun is a Linux tun/tap device (tun mode)                                                                                                                                                                                                                              
Listening on 0.0.0.0 port 655                                                                                                          
Listening on :: port 655                                                                                                                                                                                                                                                       
Ready                                                                                                                                  
Trying to connect to [redacted] ([redacted] port 20)                                                                                                                                                                                                                              
Connected to [redacted] ([redacted] port 20)                                                                                              
Sending ID to [redacted] ([redacted] port 20): 0 [redacted] 17.7                                                                          
Sending 18 bytes of raw metadata to [redacted] ([redacted] port 20)                                                                                                                                                                                                               
[...]

PR #273, opened 2021-06-29
Fix stack smash

splitice:
Issue #252 proposed fix.

1. Prevent overrun of `send_udp_probe_packet`. It's low cost and helps make tinc safer for the future.
2. Prevent negative interval

In the case of negative interval minmtu will be used.


gsliepen:
Cherry-picked and reformatted as 9eb34ebfbe763e50b6efb1168dec52331559f7d1, e7a422c64423c3fdb4d44235440ea8dd0ba1d9f5 and bceaa804d57ac88f5201886d48738975ee2ae8c2.

Issue #266, opened 2021-06-25
Experimental: epoll support


splitice:
A follow on from issue #110 

Formed from https://www.cs.cornell.edu/courses/cs5413/2014fa/projects/group_of_ej222_jj329_jsh263/ for the purposes of evaluating the real world performance differences against a current version tinc 1.1

I thought I would share this branch https://github.com/X4BNet/tinc/tree/experiment/paper2

Still a WIP and not mergable (hence no PR).

splitice:
Current notes:
 - This removes pselect / select support (no windows)
 - This uses GNU vasprintf (no windows, no non-linux)
 - double iteration in splay tree far from optimal
 - comment removal necessary
 - committed testing script from paper
 - some debug comments committed

I'll say this again, this is intended only for evaluation at this point. Under no circumstances run this fork in production.

splitice:
OS event support:
 - BSD: kqueue / **pselect** / select
 - Android: **epoll** / pselect / select
 - Linux: **epoll** / pselect / select
 - Windows: **IOCP** / select

Bold is the intended method in my fork.

splitice:
Double iteration can be resolved by supplying a user pointer to the event object. You can supply up to 64bits of data that the kernel will provide back to the caller of epoll_wait.

This could be a pointer to to the `io_t` perhaps? Although I'm not yet sure how the splay tree works and the safety requirements of storing the pointer.

splitice:
I benchmarked `experiment/paper2` against the latest `1.1` branch.

The experiment branch was consistently slightly lower resource (higher performing). Around 5-7% faster depending on the test run. Performance improvement is mostly in "system" cpu usage (8%) with a small amount of user CPU usage (3%).

Testing was carried out with approx 17 k PPS in both directions (high rate ICMP ping from hping3) between DigitalOcean SFO3 and Vultr LA. Speed of 17k PPS was chosen as it is the upper rate possible with tinc on this hardware before major VPN instability occurs.

This performance improvement would likely decrease with larger packets as encryption (approx 10 - 15% of total CPU usage on small packets in my experience) would become a larger percentage.

splitice:
Idle usage is also lower (25%). Not too important though. Likely just because of the smaller syscalls (select vs epoll).

splitice:
@derekchiang @jasdeep-hundal

The code attached to your paper appears to have been licenced under GPL. Could you please confirm, or if willing make a release into public domain.

splitice:
[...]

PR #275, opened 2021-07-02
initial epoll & initial tx buffering support

splitice:
adds epoll and basic tx buffers as per research paper https://www.cs.cornell.edu/courses/cs5413/2014fa/projects/group_of_ej222_jj329_jsh263/ as developed in #266


splitice:
Packet buffers are as defined in the paper. They arent particularly useful yet outside of broacasted packets.

Tun ARP messages would probably be the main user of the buffering capability?

gsliepen:
> adds epoll and basic tx buffers

It would be nice to split this into two pull requests, one for `epoll` support and one for `sendmmsg()`, as I think they are orthogonal.

> Packet buffers are as defined in the paper. They arent particularly useful yet outside of broadcasted packets.
> 
> Tun ARP messages would probably be the main user of the buffering capability?

Why not use it for all UDP packets? Just always add packets to the socket's `packet_buffer`, and then call `send_buffered_packets()` if either the packet buffer is full, or right before calling `select()` or `epoll_wait()`.

splitice:
As per the issue larger architectual issues prevent that currently. All fds would need to be non-blocking.

I'm not super against splitting this however I feel they both support each other and would likely conflict if your suggestion was implemented.

gsliepen:
All fds are non-blocking. Also, I still don't see how the sendmmsg and epoll parts depend on each other. There is nothing in your code that suggests so.

splitice:
> and then call send_buffered_packets() if either the packet buffer is full, or right before calling select() or epoll_wait().

Introduces the expected cross dependency mentioned. I'd consider this packet buffering functional but basic currently. It's greatly extended in the vhost net work (which also depends on epoll_wait).

Unfortunately I have limited time available so want to merge as much as possible where possible to keep conflicts to a bare minimum.

---

We discussed making the loop safe to handle multiple io operations per loop, I'd prefer if you were to make that commit. At this stage I don't feel confident enough in my undestanding of that area.

splitice:
Since hashtable has merged I'll start work on getting this up to merge.

Removed the splay each iteration of the io tree, this should boost performance on the epoll path. Needs further testing though.

I'm looking into the multiple IO ops through the event paths.

splitice:
It would appear that epoll_wait is more efficient than doing multiple read/writes per cycle even at moderate rates of transfer (and significantly better at low rates). The 2x syscall for every socket quickly dwarfs gains from reduction in epoll_waits.

splitice:
I'm thinking I'll submit a seperate PR at around a0ff2d874b39959066994b5486807fdd7a638e40 with just the epoll parts.

I think a more general cleanup and refactor of net_packet as discussed in #319 would be better done before packet buffers are added (as nice as they are).

splitice:
Closing PR. Will resubmit packet buffers at a later date. Epoll V2 pr already sent.

PR #280, opened 2021-07-10
GitHub actions cleanups; running tests on Windows

hg:
Split off #277 for easier review. This currently won't pass CI checks until both #277 and #279 are merged (with this rebased on top) because of the compatibility fixes introduced there.

---

These are some things here that could have been done better the first time (like actually testing .debs before publishing them). Sorry about that!

  - check code formatting only once
  - check test scripts formatting (shfmt)
  - static analysis for test scripts (shellcheck)
  - save more test logs and other debug info
  - add missing pieces to Ubuntu packages
  - test .debs on clean machine before publishing ([example](https://github.com/hg/tinc/releases/tag/release-1.95))
  - git clone full history for changelog generation
  - rename some steps


.github/workflows/test.yml
```diff
@@ -29,60 +29,83 @@ jobs:
...
-    runs-on: ${{ matrix.os }}
-    env:
-      CC: ${{ matrix.compiler }}
+      matrix:
+        os:
+          - alpine
+          - centos:7 # aka RHEL 7
+          - almalinux:8 # aka RHEL 8
+          - debian:oldstable
+          - debian:stable
```
gsliepen, line 108: Is there a debian:testing? If so, that would be great to add.

hg: Sure, I'll add testing and sid to the mix. I was afraid you say that there are too many images already.

gsliepen: Maybe. But Debian stable can be quite old at times, having testing is helpful. If it's too much, I'd rather keep Debian oldstable+stable+testing to ensure we can build against old and new versions of libraries, and Alpine because it uses musl and is quite different from the other distros.

gsliepen:
The changes look OK to me; they should be rebased, then we can see if they pass the tests.

hg:
Sure, I was putting some finishing touches to them. Here's the current state.

The ASAN job will be failing for the time being because of a bunch of known memory leaks. Some of them are not important (a single allocation to a global variable which is not cleaned up before the `exit()`), and some look like they are. I've been cleaning them up and suppressing others where it does not make sense to add two dozen `free()`s because we're exiting anyway, but it'll take some time.

Linux jobs now do two test runs — with and without legacy protocol support. I wanted to add a libgcrypt test run, but couldn't get it to link without also adding OpenSSL (it requires some functions that are only defined in `openssl/cipher.c` and `openssl/rsa.c`).

Windows and macOS should still be split in two separate jobs IMO, because if one of them fails, you still have the other one as reference. We can add more jobs, but they run much slower than Linux, and definitely have much lower concurrency limits (Mac jobs cost 10× as much as Linux jobs for paying customers!)

gsliepen:
Ok, lots of missing calls to `free()` at program exit time making the address sanitizer disappointed. I'll fix those first.

gsliepen:
Ah, I only just now saw your comment. If you are going to fix them, be my guest.

hg:
Nah, I'll be happy to do it. Most of them are already fixed [here](https://github.com/hg/tinc/tree/memleaks), I just need to find a way to disable `vasprintf` in CI without patching tinc (otherwise you get very short stack traces whenever it produces a leak, and those are impossible to suppress), and recheck everything thrice.

gsliepen:
Alternatively, remove testing for `--with-libgcrypt` and the address sanitizer from this pull request, and add it back together with the necessary fixes in another pull request. But it's up to you :)

hg:
This is probably the best solution. Cleaning up those leaks will take some time, and everything else here is ready.

[Demo](https://github.com/hg/tinc/actions/runs/1046312226).

ASAN can be re-enabled by adding `address` to `jobs.sanitizer.strategy.matrix.sanitizer` in `test.yml`:

```yaml
      matrix:
        sanitizer:
          - address
          - thread
          - undefined
```

Alpine is still failing a bit more often than I would like, but it will be impossible to fix quickly with my current level of knowledge of tinc's internals. Just don't pay too much attention to it for now.

---

By the way, UBSAN [produces a lot](https://github.com/hg/tinc/suites/3276867144/artifacts/76365241) of these things:

```
net_setup.c:234:17: runtime error: implicit conversion from type 'int' of value -33217 (32-bit, signed) to type 'unsigned int' changed the value to 4294934079 (32-bit, unsigned)
    #0 0x55be8d73e42a in read_ecdsa_private_key /home/runner/work/tinc/tinc/src/net_setup.c:234:17
    #1 0x55be8d73b3c0 in setup_myself /home/runner/work/tinc/tinc/src/net_setup.c:866:18
    #2 0x55be8d73ab54 in setup_network /home/runner/work/tinc/tinc/src/net_setup.c:1246:6
    #3 0x55be8d7abf26 in main /home/runner/work/tinc/tinc/src/tincd.c:513:6
    #4 0x7f0f6801e0b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)
    #5 0x55be8d6c4dfd in _start (/home/runner/work/tinc/tinc/src/tincd+0x9ddfd)
```

This is easy enough to fix IIUC:

```diff
diff --git a/src/net_setup.c b/src/net_setup.c
@@ -231,7 +231,7 @@ static bool read_ecdsa_private_key(void) {
 		return false;
 	}

-	if(s.st_mode & ~0100700) {
+	if(s.st_mode & ~0100700u) {
 		logger(DEBUG_ALWAYS, LOG_WARNING, "Warning: insecure file permissions for Ed25519 private key file `%s'!", fname);
 	}

```

and I am willing to do the work just for the higher signal-to-noise ratio. Would you be willing to accept these changes, or do we go the suppression route?

Thanks!

gsliepen:
I'll accept those changes.

hg:
Additional BSD checks

- https://builds.sr.ht/~reducer/job/548869
- https://builds.sr.ht/~reducer/job/548868
- https://builds.sr.ht/~reducer/job/548867

PR #285, opened 2021-07-16
Allow running sptps_test on Windows

hg:
`sptps_test.c` was hanging on Windows and never sending any data because one is not supposed to call `select()` there on anything except proper BSD sockets.

After looking around the web for possible solutions I wasn't able to find anything less hackish than creating a separate thread which reads stdin and pushes data through a TCP socket to the main thread. The main thread can reuse the same `select()` loop that's used for every other operating system.

Here are the test results:
- [one](https://github.com/hg/tinc/runs/3087517825?check_suite_focus=true#step:6:56)
- [two](https://github.com/hg/tinc/runs/3087517790?check_suite_focus=true#step:6:57)
- [three](https://github.com/hg/tinc/runs/3091850792?check_suite_focus=true)
- [four](https://github.com/hg/tinc/runs/3091850806?check_suite_focus=true)
- [five](https://github.com/hg/tinc/runs/3091909047?check_suite_focus=true)
- [six](https://github.com/hg/tinc/runs/3091909055?check_suite_focus=true)
- [seven](https://github.com/hg/tinc/runs/3091952703?check_suite_focus=true)
- [eight](https://github.com/hg/tinc/runs/3091952707?check_suite_focus=true)
- [nine](https://github.com/hg/tinc/runs/3091983584?check_suite_focus=true)
- [ten](https://github.com/hg/tinc/runs/3091983594?check_suite_focus=true)

My C skills are very rusty. Apologies for the eye bleeds.

### DIFF
--- a/src/sptps_test.c
+++ b/src/sptps_test.c
@@ -25,10 +25,22 @@
 
 #include <getopt.h>
 
+#ifdef HAVE_MINGW
+#include <pthread.h>
+#endif
+
 #include "crypto.h"
 #include "ecdsa.h"
 #include "sptps.h"
 #include "utils.h"
+
+#ifndef MAX
+#define MAX(a, b) ((a) > (b) ? (a) : (b))
+#endif
+
+#ifndef HAVE_MINGW
+#define closesocket(s) close(s)
+#endif
 
 // Symbols necessary to link with logger.o
 bool send_request(void *c, const char *msg, ...) {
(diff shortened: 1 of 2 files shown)


src/sptps_test.c
```diff
@@ -125,6 +134,149 @@ static void usage() {
 	fprintf(stderr, "Report bugs to tinc@tinc-vpn.org.\n");
 }
 
+#ifdef HAVE_MINGW
+
+// Windows does not allow calling select() on anything but sockets. Therefore,
```
gsliepen, line 139: The proper way on Windows is to use `WSAWaitForMultipleEvents()`, and there is some way to have that work with BSD sockets. This is what's done in `event.c`. It might actually be less code to go that way than to have a separate thread and TCP socket. But we can also add that later.

src/sptps_test.c
```diff
@@ -125,6 +134,149 @@ static void usage() {
...
+			}
+
+			if(verbose) {
+				fprintf(stderr, "Sent %lld bytes to peer\n", nsend);
+			}
+		}
+
+		shutdown(peer_fd, SD_SEND);
+
+		while(recv(peer_fd, buf, sizeof(buf), 0) > 0) ;
```
gsliepen, line 183: Why do we need this? This socket should never receive anything to begin with.

src/sptps_test.c
```diff
@@ -125,6 +134,149 @@ static void usage() {
...
+
+		closesocket(peer_fd);
+	}
+
+	shutdown(*sock_fd, SHUT_RDWR);
```
gsliepen, line 188: The call to `shutdown()` isn't necessary.

src/sptps_test.c
```diff
@@ -125,6 +134,149 @@ static void usage() {
...
+		ssize_t nread;
+
+		while((nread = read(STDIN_FILENO, buf, sizeof(buf))) > 0) {
+			if(verbose) {
+				fprintf(stderr, "Read %lld bytes from input\n", nread);
+			}
+
+			ssize_t nsend = send(peer_fd, buf, nread, 0);
+
+			if(nsend <= 0) {
```
gsliepen, line 171: `send()` might send less bytes than you request. You have to handle this by adding a loop around it that send the remaining bytes if necessary. Also, if it returns -1, then this might still not be a fatal error. Use:

```
if (nsend == -1) {
    if (sockwouldblock(sockerrno)) {
        /* retry */
    } else {
        /* it really is an error, abort */
    }
}
```

src/sptps_test.c
```diff
@@ -125,6 +134,149 @@ static void usage() {
...
+	}
+
+	shutdown(*sock_fd, SHUT_RDWR);
+	closesocket(*sock_fd);
+	free(sock_fd);
+}
+
+int start_input_reader() {
+	// Get a random port in range 20000..50000
+	int port = rand() % 30001 + 20000;
```
gsliepen, line 195: You can bind to port 0, this will let the operating system choose a port. You can then get the address+port you need to `connect()` to it using `getsockname()`.

src/sptps_test.c
```diff
@@ -125,6 +134,149 @@ static void usage() {
...
+		return -1;
+	}
+
+	if(listen(sock_fd, 1) < 0) {
+		fprintf(stderr, "Could not listen: %s\n", strerror(errno));
+		closesocket(sock_fd);
+		return -1;
+	}
+
+	int *server_fd = malloc(sizeof(*server_fd));
```
gsliepen, line 232: That's a bit overkill. I would just store it in a global variable, then there is also no need to pass a pointer to it to  `pthread_create()`.

hg:
Oh-oh. Thank you for the review and for spoonfeding me. Everything should be fixed. I'll look into implementing the proper solution with `WSAWaitForMultipleEvents` later. If you decide against merging this, it only fixes one test on a single operating system. We can just disable the test for now.

[Checks](https://github.com/hg/tinc/actions/runs/1040302363).

---

FWIW I think #277 is ready for merging. It's marked as work in progress as an additional safety measure because it contains a lot of junk that needs to be dropped at the last possible moment.

There's no low hanging fruit left in there. I know of at least one bug that crops up periodically (tinc-to-tincd connection failure, more frequent in musl builds and on Windows), but it looks like a bug in tinc's internals. The whole test suite has been running for ~80 hours on my aarch64 machine before it happened, so it needs some careful debugging.

---

By the way, while looking for build instructions for Alpine I noticed [this](https://github.com/alpinelinux/aports/blob/d424fe9e3953ec7b193c613803b17af165229b7a/community/tinc-pre/APKBUILD#L9-L11)

```sh
# s390x: tests hang
# armhf: tests fail
arch="all !s390x !armhf !armv7"
```

and [this](https://github.com/alpinelinux/aports/blob/d424fe9e3953ec7b193c613803b17af165229b7a/community/tinc-pre/APKBUILD#L51-L52)

```sh
	# Disable test
	sed -i '/legacy-protocol.test/d' test/Makefile
```

so having tests run on more architectures would definitely be nice.

PR #286, opened 2021-07-18
RFC: unit testing for C

hg:
Note: this is purely an RFC at this point.

cc @splitice

---

There was some interest in adding C unit testing to the project.

There doesn't seem to be an agreement about any de facto standard, so this PR adds two well maintaned, popular, and tested libraries for your evaluation. Pick something (or suggest something else) and I'll remove the rest.

Some popular libraries were not considered because of their reliance on a different build system or a working C++ compiler (and the the rest is _really_ similar to cmocka).

The choice here is between:

- a very lightweight library with can be fully bundled with the project;
- or a bit more features which bring a separate dependency (with different versions on every platform, or the requirement of a separate source build).

# Unity

[Link](https://github.com/ThrowTheSwitch/Unity).

See sample tests in [unittest/unity](https://github.com/hg/tinc/tree/unit-tests/unittest/unity).

How to use (remove other test section from `unittest/Makefile.am` or install the appropriate library beforehand):

    $ make check

## Pros

- the simplest solution, basically just a bunch of convenience macros that you would have to write anyway.
- can be easily vendored (which I did), and does not require installing anything on dev machines or CI systems.

## Cons

- all tests from the same source file run as a single process — a misbehaving test can corrupt the whole group.
- no support for test mocks, this requires [another library](https://github.com/ThrowTheSwitch/CMock/) which pulls Ruby and a bunch of other things.
- single output format (not really important at the moment, but would be nice to have if we were to add some sort of autogenerated summary in the future).

## Similar libraries

- https://github.com/silentbicycle/greatest
- https://github.com/siu/minunit
- https://github.com/nemequ/munit

# cmocka

[Link](https://api.cmocka.org).

See sample tests in [unittest/cmocka](https://github.com/hg/tinc/tree/unit-tests/unittest/cmocka).

More realistic tests from other projects: [SSSD](https://github.com/SSSD/sssd/tree/master/src/tests/cmocka), [samba](https://github.com/samba-team/samba/search?q=cmocka), [OpenVPN](https://github.com/OpenVPN/openvpn/search?l=C&q=cmocka), [coreboot](https://github.com/coreboot/coreboot/search?l=C&q=cmocka).

How to use:

    # apt install libcmocka-dev
    $ make check

## Pros

- supports mocking out of the box (see more [here](https://lwn.net/Articles/558106/) and [here](https://gitlab.com/cmocka/cmocka/-/blob/master/example/mock/chef_wrap/waiter_test_wrap.c)).
- does its own memory corruption checks which could be helpful on [platforms](https://github.com/google/sanitizers/wiki/AddressSanitizer) where ASAN is not available.
- many output formats, should we require that at some point.

## Cons

- requires installing a pre-built library on every dev machine and every CI build, or building from source.
- same limitation about the single process.

The first one is a real pain. The tests that ran just fine on my machine [failed on Ubuntu 18.04](https://github.com/hg/tinc/runs/3097291598) because its version of cmocka does not provide the newer functions that were used.

## Similar libraries

- https://libcheck.github.io/check/
- https://github.com/Snaipe/Criterion
- https://gitlab.com/cunity/cunit/

---

PS: don't pay attention to the Makefile changes. Let me know which type of a testing library you'd rather see and I'll redo them from scratch (probably using libtool).


gsliepen:
Nice. I think depending on cmocka is fine. It certainly allows us to do more kinds of tests, like the one mocking the `execute_script()` function. We don't need to do unit testing on all platforms; 95% of the code is platform independent, and I think the rest is covered by the integration tests.

hg:
Alright, thank you. I'll rework this in the next few days when time permits.

I am not too sure about the best way to link tests against the project's code. Which solution would you rather see?

1. as it's implemented currently: build `tinc` / `tincd` (except the `main()` object files) into two separate static libraries, and then:
    - link each library against its `main()` object file to produce the binary
    - link one of the libraries against each test

This will probably require switching the project to libtool because the libraries it produces can remember the required linker flags (no need to write `-lminiupnpc` twice) ([link](https://www.gnu.org/software/automake/manual/html_node/Libtool-Convenience-Libraries.html)).

2. move tests inside `src/tests` and add them to `src/Makefile.am` (no need for libraries then, just pass dependencies in variables)

3. remove `src/Makefile.am` and `unittest/Makefile.am` and move their contents to the top `Makefile.am` (same)

I've also been [experimenting](https://github.com/hg/tinc/blob/meson/meson.build) with porting tinc to the meson build system. With meson all of this would be way easier to implement. But this is probably asking too much for now.

gsliepen:
All options are fine, perhaps option 2 is the cleanest of them all.

Moving to Meson would be very nice as well; I've been using that in other projects, but I didn't do this for tinc because it's not so trivial; there's conditional building of source files, and the test suite might need some work as well to work with meson. Last but not least, autotools provides *excellent* support for cross-compiling out of the box, and I've been using that to build the Windows binaries on Linux. With Meson you need to write "cross files" yourself, and I wonder how portable they are.

splitice:
@hg cmocka looks awesome.

Is it worth exploring some abstraction to split fixtures into seperate processes?

lancethepants:
Sorry, off topic for this pull request, but since this is where the discussion is happening.

+1 for staying with autotools, hands down the best build system for cross-compiling.
```./configure --host=arm-linux```

I'd even take cmake before meson.
```
cmake \
-DCMAKE_CROSSCOMPILING=True \
-DCMAKE_C_COMPILER="arm-linux-gcc" \
-DCMAKE_CXX_COMPILER="arm-linux-g++" \
.
```
https://mesonbuild.com/Cross-compilation.html
`Meson requires you to write a cross build definition file. `

If you can't configure a cross-compile in a single command then cross-compilation is a 2nd class citizen.

libGLib moved to meson, and after tinkering with it for two hours I decided I was wasting my time and am staying with their last autotools build until I don't have anything better to do then mess with cross-compiling with meson.

Tinc runs on a host of architectures. I myself run it on mipsel, arm, aarch64, android, x86_64 on windows and linux.  I even compile it myself for all these target minus android. Complicating the build process will raise the barrier to entry for those wanting to compile and try it out.

I applaud tinc for its simplicity. I integrated tinc+gui into tomato firmware, and it's been a hit!  People love the simplicity of easily creating a mesh vpn.  Let's please keep the build process simple as well.

**Comment by hg** [comment]
@lancethepants thanks for the input! I do not speak for the project, but if this complicates cross-compilation so much, I definitely would not want to break it. I'll put the idea on the back-burner for the time being, maybe the situation will improve some day.

**Comment by hg** [comment]
After trying several approaches, I think shoving everything into one giant makefile is the best autotools can offer.

If you're curious, my libtool experiments can be found [here](https://github.com/hg/tinc/tree/cmocka-libtool). [Convenience libraries](https://www.gnu.org/software/automake/manual/html_node/Libtool-Convenience-Libraries.html) (which is what seems to be the preferred solution for code reuse) proved to be more trouble than they're worth — you get the separation between project sources and test sources, plus a proper top-level unit test directory, but pay with:

- five additional m4 macro files
- ~50% more compilation time (because it needs to build most sources twice — once for the two libtool libraries, and once for `sptps_*` binaries; the blame [doesn't seem](https://www.gnu.org/software/automake/manual/html_node/Objects-created-both-with-libtool-and-without.html) to lie entirely on my incompetence)
- [weird](https://github.com/hg/tinc/runs/3126726806) runtime [errors](https://github.com/hg/tinc/runs/3127013282), particularly [on Windows](https://github.com/hg/tinc/runs/3127733880) (this can be worked around as [described here](https://www.openssl.org/docs/faq.html) (look for `compiled a program under Windows and it crashes`), but who knows what else is broken and won't be found until it hits end-user machines)

What's nice is that it remembers linker flags and does not require you to write three lines of Makefile boilerplate for every new test you add.

Some projects (e.g. [profanity](https://github.com/profanity-im/profanity/blob/master/Makefile.am)) bundle all unit tests into a single binary, but [this](https://github.com/profanity-im/profanity/blob/master/tests/unittests/unittests.c) is pretty difficult to read IMO.

Would you rather have everything thrown into a single test binary (so we get faster compilation times and smaller makefiles), or do we keep them separated?

---

Run:

    $ make -j check

Autoconf will perform a couple of checks:

- if `libcmocka` is not installed, unit testing will be skipped altogether
- if your linker does not support function wrapping (`--wrap`), tests that require mock objects to run will be skipped with a warning

Tests:

- https://github.com/hg/tinc/actions/runs/1056085728
- https://github.com/hg/tinc/actions/runs/1056833176

---

> Is it worth exploring some abstraction to split fixtures into seperate processes?

I wouldn't know of any decent abstractions right now. If you have any ideas, please do share, I'm willing to do the work.

File: src/tincctl.c [hunk 1]
```diff
@@ -57,27 +57,28 @@ static bool show_version = false;
 
 static char *name = NULL;
 static char controlcookie[1025];
-char *tinc_conf = NULL;
-char *hosts_dir = NULL;
-struct timeval now;
+extern char *tinc_conf;
```
gsliepen, line 60: Maybe it's better to have a `global.h` where we declare all global variables.

src/Makefile.am
```diff
@@ -273,24 +273,53 @@ sptps_speed_SOURCES += \
...
+if HAVE_CMOCKA
+## Use TAP log driver (gives each test result separately in `make check` output)
+LOG_DRIVER = env AM_TAP_AWK='$(AWK)' $(SHELL) $(top_srcdir)/tap-driver.sh
+
+AM_TESTS_ENVIRONMENT = \
+	export CMOCKA_MESSAGE_OUTPUT=TAP;
+
+AM_CFLAGS += -iquote unittest
+
+tinc_test_LDADD = $(tinc_LDADD) -lcmocka
```
gsliepen, line 316: This causes a warning from automake that there is no binary `tinc_test`. Changing the `_LDADD` suffix to `_ldadd` fixes it.

unittest
```diff
@@ -0,0 +1 @@
+src/unittest/
```
gsliepen, line 1: This symlink doesn't seem necessary?

src/unittest/Makefile.am
```diff
@@ -0,0 +1,19 @@
...
+##     target_LDFLAGS = $(tincd_LDFLAGS)
+##     target_LDADD   = $(tincd_test_LDADD)
+## To link against tinc, replace `tincd` with `tinc` in the directives above.
+## Paths below are relative to src/
+
+TESTS += \
+	test_subnet \
+	test_splay_tree
+
+test_subnet_SOURCES = $(tincd_common_SOURCES) unittest/test_subnet.c
```
gsliepen, line 13: The sources must also list the header files that are used, otherwise those are not copied into the tarball when running `make dist`.

src/unittest/test_net.c
```diff
@@ -0,0 +1,45 @@
+#include "unittest.h"
+#include "net.h"
```
gsliepen, line 2: The `#include` directives with quotes are relative to the source file, so this should be `#include "../net.h"`, same for other .h files from the parent directory.

src/Makefile.am
```diff
@@ -273,24 +273,53 @@ sptps_speed_SOURCES += \
...
 AM_CPPFLAGS = $(CODE_COVERAGE_CPPFLAGS)
+
+if HAVE_CMOCKA
+## Use TAP log driver (gives each test result separately in `make check` output)
+LOG_DRIVER = env AM_TAP_AWK='$(AWK)' $(SHELL) $(top_srcdir)/tap-driver.sh
+
+AM_TESTS_ENVIRONMENT = \
+	export CMOCKA_MESSAGE_OUTPUT=TAP;
+
+AM_CFLAGS += -iquote unittest
```
gsliepen, line 314: This whole line should be removed, instead proper relative paths should be used in `#include` directives.

gsliepen (changes requested):
This looks mostly fine, but it currently fails `make distcheck`.

eli-schwartz:
> Moving to Meson would be very nice as well; I've been using that in other projects, but I didn't do this for tinc because it's not so trivial; there's conditional building of source files, and the test suite might need some work as well to work with meson. Last but not least, autotools provides *excellent* support for cross-compiling out of the box, and I've been using that to build the Windows binaries on Linux. With Meson you need to write "cross files" yourself, and I wonder how portable they are.

Conditional building of source files should be as easy to do in meson as it is in autotools.

For cross compiling, you may be able to get away with this:

```
CC=/path/to/my/$TRIPLET-gcc meson setup builddir/ --cross-file /dev/null
```

meson will trigger cross building due to passing the cross-file option, but no settings will be added because the file is an empty file. Meanwhile, CC is respected as usual and sets the cross compiler you want.

Most of the actual settings you *can* set, probably don't *need* to be set.

...

In terms of cross file portability, what is supposed to be portable -- or not portable -- about them? The purpose of a cross file is to *override* settings such as a compiler, a PKG_CONFIG_LIBDIR, and various other things. Anything interesting to set, is probably heavily dependent on your local system. This is not portable on autotools either, but autotools uses custom user-specific environment variables to perform that configuration, so it looks less intimidating?

Meson cross files can have a `[constants]` section and refer to it in other values, so you can have each user specify e.g. the root directory of a cross toolchain, and then share a portable cross file that does `cpp = basedir / 'aarch64-linux-gnu-g++'`.

Cross files can be layered. Specify two of them, and one can be a user-supplied set of constants while the other is a portable toolchain file.

hg:
@gsliepen

Apologies, I've been a bit busy lately. Please have another look.

[tests](https://github.com/hg/tinc/actions/runs/1970377236) (two jobs ran into timeout because of slow disk i/o)

---

It makes senses to not bother to support cmocka older than 1.1.5. I was sitting on this PR for so long it's now available pretty much everywhere.

Not sure about the comment on `global.h`, though. Should *all* `extern` variable declarations be moved into separate headers (one for `tinc` and one for `tincd`)?

Also not sure about `PKG_CHECK_MODULES`. It adds a new dependency on `pkg-config` which is not really needed for anything else.

Which solution you'd prefer — require `pkg-config` to run unit tests, or leave the current version in place?

Here are rough drafts for both:

# `pkg-config`

```m4
dnl Check the existence of cmocka unit testing library

AC_DEFUN([tinc_CMOCKA], [

  PKG_CHECK_MODULES([CMOCKA], [cmocka >= 1.1.5], [
    cmocka_found=yes

    AX_CHECK_LINK_FLAG([-Wl,--wrap=func], [
      linker_has_wrap=yes
      AC_DEFINE(HAVE_MOCK_SUPPORT, 1, [linker has support for the --wrap flag.])
    ], [
      AC_MSG_WARN("linker does not support function wrapping; tests that require mocks will not run.")
    ], [-Werror])

  ], [
    AC_MSG_WARN("cmocka was not found; unit tests will not run.")
  ])

  AM_CONDITIONAL(HAVE_CMOCKA, test "x$cmocka_found" = 'xyes')

  # can use --wrap for test mocks
  AM_CONDITIONAL(HAVE_MOCK_SUPPORT, test "x$linker_has_wrap" = 'xyes')

])
```

# manual checks

```m4
dnl Check the existence of cmocka unit testing library

AC_DEFUN([tinc_CMOCKA], [

  m4_define([cmocka_headers], [
    #include <stdarg.h>
    #include <stddef.h>
    #include <setjmp.h>
    #include <cmocka.h>
  ])

  dnl Neither AC_CHECK_LIB nor AC_CHECK_FUNC allow adding custom headers.
  AC_CHECK_DECL(assert_int_equal, [
    cmocka_found=yes

    AX_CHECK_LINK_FLAG([-Wl,--wrap=func], [
      linker_has_wrap=yes
      AC_DEFINE(HAVE_MOCK_SUPPORT, 1, [linker has support for the --wrap flag.])
    ], [
      AC_MSG_WARN("linker does not support function wrapping; tests that require mocks will not run.")
    ], [-Werror])

  ], [], cmocka_headers)

  AS_IF([test "x$cmocka_found" != "xyes"], [
    AC_MSG_WARN("cmocka was not found; unit tests will not run.")
  ])

  AM_CONDITIONAL(HAVE_CMOCKA, test "x$cmocka_found" = 'xyes')

  # can use --wrap for test mocks
  AM_CONDITIONAL(HAVE_MOCK_SUPPORT, test "x$linker_has_wrap" = 'xyes')

])
```

eli-schwartz:
> Also not sure about `PKG_CHECK_MODULES`. It adds a new dependency on `pkg-config` which is not really needed for anything else.

The PKG_CHECK_MODULES autotools macro allows users to specify CMOCKA_LIBS and CMOCKA_CFLAGS. If you don't specify an action-if-not-found, it will print this error message:

```
checking for CMOCKA... no
configure: error: Package requirements (cmocka >= 1.1.5) were not met:

Package 'cmocka', required by 'virtual:world', not found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables CMOCKA_CFLAGS
and CMOCKA_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.
```

You can include part of that messaging, like this:

```
checking for CMOCKA... no
configure: WARNING: cmocka was not found; unit tests will not run.

Alternatively, you may set the environment variables CMOCKA_CFLAGS
and CMOCKA_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.
```

Just change this:
```m4
  ], [
    AC_MSG_WARN("cmocka was not found; unit tests will not run.")
  ])
```

to this:
```m4
  ], [
    AC_MSG_WARN([cmocka was not found; unit tests will not run.

_PKG_TEXT])
  ])
```

src/tinc_global.c
```diff
@@ -0,0 +1,18 @@
+#include "system.h"
+
```
gsliepen, line 2: This should also include `"tinc_global.h"`. I think the `tinc_` prefix for the filename is also not necessary. Last but not least, it might be nice to ensure the .h and .c file declare/define the exact same variables.

hg: The whole idea of rearranging globals was a mistake on my part. It's probably best to do this separately later.

I created simple wrappers around `main()` to avoid linker errors. This should suffice and is more straightforward.

There are some intermittent failures with really weird [errors](https://github.com/hg/tinc/runs/5523326249?check_suite_focus=true#step:11:1002). Not sure what's causing it, as it only shows in github CI. I'll look into it.

gsliepen:
It's great to have working unit tests, so I am tempted to merge this if you think it is ready. However, while I did say in the past to go for building the unit tests inside src/, as that was the simplest automake solution, it makes the src/ directory quite messy, for two reasons:

1. The need for changing the `main()` function, and adding stub tinc/tincd_main.c files. Ideally, no changes to the actual source should be needed for unit tests. But partly this is due to the way global variables are used in tinc, perhaps that requires some cleanup.
2. Test binaries end up inside src/. Switching to meson was discussed, and while @lancethepants and I thought it would be best to stick with autotools for its excellent cross-compilation support, @eli-schwartz mentioned it should be possible to make cross-compilation easy with meson too. It would also be a step towards compiling tinc natively on Windows without using MinGW, although that would bring its own challenges.

I don't know if you think it's worth adressing either of these points in this pull request or if that should be done at a later point?

eli-schwartz:
For what it's worth, I've heard from certain groups that they find Meson's cross compilation support much better in their workflow than autotools. e.g. Void Linux which actually does cross compile the entire distro.


(Of course it is better than cmake, cmake is legendarily bad at cross compiling...)

hg:
@gsliepen not sure TBH. I tried rearranging globals properly yesterday, and it turned into such a mess (touching almost every file in the project) that I thought it best to create a couple of stubs and call it a day.

Some global variables alias (they have the same name even if it's two separate globals underneath — one for tinc and one for tincd), so you get either the ugliness below, or a bunch of warnings about unnecessary redeclaration. A few examples: `now`, `scriptextension`, `scriptinterpreter`.

I think in longer term it would make sense to combine them into larger structs and then (maybe) get rid of globals altogether by passing them as some sort of context.

---

Here's another try. Some globals were only really used by a single translation unit; I moved them there. Others are more popular and were moved to `globals.c`.

Let me know what you think. It's in a separate commit, so it can be thrown away more easily.

---

Regarding meson: there was a bit of a meson [experiment](https://github.com/hg/tinc/commit/67280e127e161a5ef9eddb31c34baa997dd39c6a) last year, I'll try cross-compiling it and see how it goes.

gsliepen:
Hm, I think I'll have a try at rearranging where the global variables are. Some are just in the wrong place; like `scriptextension`, which should definitely just live in script.c and not in net_setup.c. The end goal is not to have any occurence of `extern` anymore in .c files, and ideally also no need for an actual global.c, but still allow unit tests, tinc and tincd to compile without hacks.

About the meson experiment, it looks good, there are some issues though, like the check for `res_init()` failing, paths in manpages have quote marks added incorrectly. It looks like cross-compiling does need a cross build definition file after all, since it doesn't seem to be able to deduce "windows" has the host machine's system if you tell it to use the MinGW compiler. With a definition file it manages to build the main binaries, but the test suite fails to work correctly. Interestingly, `./configure` is only marginally slower than `meson build`, and `make -j${nproc}` is just as fast as `ninja` from a cold start. `ninja` wins for rebuilds because it uses ccache automatically, which is quite nice.

eli-schwartz:
The Meson source code repository includes the file `cross/linux-mingw-w64-32bit.txt` which could be used to cross compile using the mingw-w64 compiler from Linux, it includes an exe_wrapper entry for wine so the testsuite should work too. Of course, this assumes you have a standard cross toolchain utilizing /usr/bin

You just need to run the built binary using `$MESON_EXE_WRAPPER`. See https://mesonbuild.com/Reference-manual_functions.html#test for details.

hg:
Yeah, that meson stuff was in a half-working state. I [rebased](https://github.com/hg/tinc/tree/meson-2) it on top of current 1.1 if anyone wants to have a look at it.

Unit tests are not supported yet (obviously).

I'll look at cross-compiling it a bit later.

```
$ ninja -C build test
ninja: Entering directory `build'
[112/113] Running all tests.
 1/14 executables.test             OK               0.05s
 2/14 basic.test                   OK               0.19s
 3/14 commandline.test             OK               0.51s
 4/14 invite-join.test             OK               0.48s
 5/14 invite-offline.test          OK               0.42s
 6/14 import-export.test           OK               0.82s
 7/14 invite-tinc-up.test          OK               0.43s
 8/14 variables.test               OK               0.41s
 9/14 scripts.test                 OK               1.11s
10/14 sptps-basic.test             SKIP             0.03s   exit status 77
11/14 ns-ping.test                 SKIP             0.03s   exit status 77
12/14 algorithms.test              OK               0.83s
13/14 legacy-protocol.test         OK              14.76s
14/14 security.test                OK              70.54s


Ok:                 12
Expected Fail:      0
Fail:               0
Unexpected Pass:    0
Skipped:            2
Timeout:            0
```

hg:
I was wondering about linking *everything* into a single static library, and then adding `tincctl.c` / `tincd.c` to it to produce two main binaries.

There are some conflicts, but those should be solvable by shuffling the globals around.

Should make `lib_tinc` and `lib_tincd` unnecessary, and all unit tests could be linked with the same library, simplifying their definition.

Bad idea?

gsliepen:
I wonder how that would affect the final size of the binaries.

hg:
I checked when I was adding first static libraries for faster compilation. Doesn't seem to. meson uses 'thin' static libraries which are just a bunch of object files in a directory, with a symbol table nearby. It links only those object files that are actually needed.

PR #78, opened 2015-05-10
Add LZ4 block compression support.

dajhorn:
Revision 2, optimized for speed, and converted to a builtin.

PR #287, opened 2021-07-20
Add support for the lz4 compression algorithm

hg:
This adds lz4 block compression support, using either the vendored copy of the library, or (preferably) the one already found on the system.

Most of the work was done by Darik Horn (@dajhorn) in this pull request: #78.

---

Changes:

- system lz4 is now tried first
- builtin lz4 is used as a fallback if the system one is too old or missing
  - this can be prevented with `--disable-lz4-builtin` (in which case the build fails)
  - use of the builtin copy can be forced with `--enable-lz4-builtin`, even if the system library works fine
- support for versions older than r129 has been removed, and the builtin copy is used in that case
  - it was released in May 2015
  - Debian oldstable and Ubuntu 18.04 ship with r131
  - CentOS 7 (oldest supported release) ships with 1.8.3 — much newer than r131

This behavior is documented.

Tests: [this](https://github.com/hg/tinc/actions/runs/1049019769), [this](https://github.com/hg/tinc/actions/runs/1048983413), and [this](https://github.com/hg/tinc/actions/runs/1048912393) (wrong package name on OpenSUSE and missing reformat from #280, everything else was fine).

---

af21369074ed3c79d5f9eccc7f9d3d2e34e83b1c is required for `compression.test` to determine which compression levels should work, and which should fail. I believe it's also quite useful by itself.

If you'd rather see a separate command (`tinc features` or `tinc has_feature`?), I can rewrite this. Either way, it can be split into a separate PR, but this commit will have to stay here for now, or `compression.test` will not pass.

```
$ ./src/tincd --version
tinc version 1.1pre18-23-g8aaf944f (built Jul 20 2021 13:54:00, protocol 17.7)
Features: openssl comp_lzo comp_zlib comp_lz4 legacy_protocol jumbograms miniupnpc uml vde

Copyright (C) 1998-2021 Ivo Timmermans, Guus Sliepen and others.
See the AUTHORS file for a complete list.

tinc comes with ABSOLUTELY NO WARRANTY.  This is free software,
and you are welcome to redistribute it under certain conditions;
see the file COPYING for details.
```

### DIFF
--- a/doc/tinc.texi
+++ b/doc/tinc.texi
@@ -339,6 +339,7 @@ message, and stop.
 * LibreSSL/OpenSSL::
 * zlib::
 * LZO::
+* LZ4::
 * libcurses::
 * libreadline::
 @end menu
@@ -458,6 +459,22 @@ from @url{https://www.oberhumer.com/opensource/lzo/}.  Instructions on how to co
 build and install this package are included within the package.  Please
 make sure you build development and runtime libraries (which is the
 default).
+
+
+@c ==================================================================
+@node       LZ4
+@subsection LZ4
+
+@cindex LZ4
+Another form of compression is offered using the LZ4 library.
+
+The LZ4 codec is bundled with Tinc and built-in by default as
+compression level 12.  Tinc can be linked to an external liblz4
+library by using the "--disable-lz4-builtin" configure switch.
+
+If LZ4 support is entirely disabled by passing "--disable-lz4" to the
+configure script, then the resulting binary will not work correctly
+on VPNs where LZ4 compression is used.
 
 
 @c ==================================================================
@@ -1325,7 +1342,7 @@ Fragmentation Needed or Packet too Big messages are dropped by firewalls.
 @item Compression = <@var{level}> (0)
 This option sets the level of compression used for UDP packets.
 Possible values are 0 (off), 1 (fast zlib) and any integer up to 9 (best zlib),
-10 (fast LZO) and 11 (best LZO).
+10 (fast LZO), 11 (best LZO), and 12 (LZ4).
 
 @cindex Digest
 @item Digest = <@var{digest}> (sha1)
--- a/doc/tinc.texi
+++ b/doc/tinc.texi
@@ -468,13 +468,18 @@ default).
 @cindex LZ4
 Another form of compression is offered using the LZ4 library.
 
-The LZ4 codec is bundled with Tinc and built-in by default as
-compression level 12.  Tinc can be linked to an external liblz4
-library by using the "--disable-lz4-builtin" configure switch.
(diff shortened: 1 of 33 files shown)


doc/tinc.texi
```diff
@@ -460,6 +461,27 @@ make sure you build development and runtime libraries (which is the
...
+@c ==================================================================
+@node       LZ4
+@subsection LZ4
+
+@cindex LZ4
+Another form of compression is offered using the LZ4 library.
+
+tinc has support for the LZ4 compression algorithm as compression level 12.
+
+By default, tinc will try to link to an external lz4 library. If it is not
```
gsliepen, line 473: lz4 -> LZ4

doc/tinc.texi
```diff
@@ -460,6 +461,27 @@ make sure you build development and runtime libraries (which is the
...
 
 
+@c ==================================================================
+@node       LZ4
+@subsection LZ4
+
+@cindex LZ4
+Another form of compression is offered using the LZ4 library.
+
+tinc has support for the LZ4 compression algorithm as compression level 12.
```
gsliepen, line 471: "tinc" at the start of a sentence should be written with a capital.

.github/workflows/deb/debian/compat
```diff
@@ -0,0 +1 @@
+11
```
gsliepen, line 1: I know you know these commits should be part of the pull request, since they are marked [REMOVE], but I am wondering, why include them in the first place?

test/compression.test
```diff
@@ -0,0 +1,133 @@
+#!/bin/sh
+
+# shellcheck source=testlib.sh
+. "${0%/*}/testlib.sh"
```
gsliepen, line 4: This should be `. ./testlib.sh` to ensure out-of-tree builds work. (This weird hack was actually required for out-of-tree builds when there was no `testlib.sh.in`, but only `testlib.sh`.)

gsliepen (changes requested):
Apart from the other comments, this looks good. Printing the features in the version output is fine, I don't see a need to make this a command for the CLI.

.github/workflows/deb/debian/compat
hg: Ah, sorry. This was cherry-picked en masse from #280 to show that this builds and runs fine on as many distributions as possible. It's just a copy of the commit there, I'll remove it completely.

Tests (these are not testing the new functionality anymore, though): [one](https://github.com/hg/tinc/actions/runs/1049291632), [two](https://github.com/hg/tinc/actions/runs/1049305258), [three](https://github.com/hg/tinc/actions/runs/1049321844)

gsliepen:
Rebased & merged. Thanks!

PR #291, opened 2021-07-25
Fix known UndefinedBehaviorSanitizer warnings plus a few more leaks

hg:
This fixes all UBSAN warnings triggered by the integration test suite, plus a bunch of leaks across the project.

`tinc fsck` has a couple more memory safety issues and some other errors (that are not currently detected by tests), but it's so complicated that I've been refactoring and testing it for a while.

It's a relatively large patch, so I'll send it separately (after this one, because it builds on top of this.)

Tests:

- https://github.com/hg/tinc/actions/runs/1064499905 (two separate successful runs that GitHub shows as a single one)
- https://builds.sr.ht/~reducer/job/551897
- https://builds.sr.ht/~reducer/job/551898
- https://builds.sr.ht/~reducer/job/551899

---

What are your thoughts on fixing *a lot* of clang-tidy warnings similar to:

- Narrowing conversion from 'unsigned long' to signed type 'int' is implementation-defined
  - these in particular often tend to trigger UBSAN at run time
- 'sscanf' used to convert a string to an integer value, but function will not report conversion errors; consider using 'strtol' instead
- 'atoi' used to convert a string to an integer value, but function will not report conversion errors; consider using 'strtol' instead
- Unused `header_name.h`
- Declaration shadows a local variable
- Declaration shadows a variable in the global scope
- …

and then adding clang-tidy with `-Werror` to CI?

I would like to reduce the number of these to zero.  However, it will produce a large diff which would be difficult to slice into pieces because fixing one warning often requires changing global functions or fields in widely used structs, which produces further warnings, and so on.

Will you accept patches to clean this up, seeing that this is not absolutely necessary and (as always) has a chance of breaking something?

### DIFF
--- a/src/hash.c
+++ b/src/hash.c
@@ -29,11 +29,11 @@ static uint32_t hash_function(const void *p, size_t len) {
 	uint32_t hash = 0;
 
 	while(true) {
-		for(int i = len > 4 ? 4 : len; --i;) {
+		for(size_t i = len > 4 ? 4 : len; --i;) {
 			hash += (uint32_t)q[len - i] << (8 * i);
 		}
 
-		hash *= 0x9e370001UL; // Golden ratio prime.
+		hash = (uint32_t)(hash * 0x9e370001UL); // Golden ratio prime.
 
 		if(len <= 4) {
 			break;
--- a/src/logger.c
+++ b/src/logger.c
@@ -94,7 +94,7 @@ static void real_logger(int level, int priority, const char *message) {
 		}
 	}
 
-	if(logcontrol) {
+	if(logcontrol && connection_list) {
 		suppress = true;
 		logcontrol = false;
 
--- a/src/net_setup.c
+++ b/src/net_setup.c
@@ -231,7 +231,7 @@ static bool read_ecdsa_private_key(void) {
 		return false;
 	}
 
-	if(s.st_mode & ~0100700) {
+	if(s.st_mode & ~0100700u) {
 		logger(DEBUG_ALWAYS, LOG_WARNING, "Warning: insecure file permissions for Ed25519 private key file `%s'!", fname);
 	}
 
@@ -323,7 +323,7 @@ static bool read_rsa_private_key(void) {
 		return false;
 	}
 
-	if(s.st_mode & ~0100700) {
+	if(s.st_mode & ~0100700u) {
 		logger(DEBUG_ALWAYS, LOG_WARNING, "Warning: insecure file permissions for RSA private key file `%s'!", fname);
 	}
 
--- a/src/net_setup.c
+++ b/src/net_setup.c
@@ -52,10 +52,11 @@ static io_t device_io;
 devops_t devops;
 bool device_standby = false;
 
-char *proxyhost;
-char *proxyport;
-char *proxyuser;
-char *proxypass;
+char *proxyhost = NULL;
+char *proxyport = NULL;
+char *proxyuser = NULL;
+char *proxypass = NULL;
+
 proxytype_t proxytype;
 bool autoconnect;
 bool disablebuggypeers;
@@ -435,18 +436,10 @@ char *get_name(void) {
 }
 
 bool setup_myself_reloadable(void) {
-	char *proxy = NULL;
-	char *rmode = NULL;
-	char *fmode = NULL;
-	char *bmode = NULL;
-	char *afname = NULL;
-	char *space;
-	bool choice;
-
 	free(scriptinterpreter);
 	scriptinterpreter = NULL;
-	get_config_string(lookup_config(config_tree, "ScriptsInterpreter"), &scriptinterpreter);
 
+	get_config_string(lookup_config(config_tree, "ScriptsInterpreter"), &scriptinterpreter);
 
 	free(scriptextension);
 
@@ -454,9 +447,13 @@ bool setup_myself_reloadable(void) {
 		scriptextension = xstrdup("");
 	}
 
+	char *proxy = NULL;
+
 	get_config_string(lookup_config(config_tree, "Proxy"), &proxy);
 
 	if(proxy) {
+		char *space;
+
 		if((space = strchr(proxy, ' '))) {
 			*space++ = 0;
 		}
@@ -475,8 +472,21 @@ bool setup_myself_reloadable(void) {
 			proxytype = PROXY_EXEC;
 		} else {
 			logger(DEBUG_ALWAYS, LOG_ERR, "Unknown proxy type %s!", proxy);
+			free(proxy);
 			return false;
 		}
+
+		free(proxyhost);
+		proxyhost = NULL;
+
+		free(proxyport);
+		proxyport = NULL;
+
+		free(proxyuser);
+		proxyuser = NULL;
+
+		free(proxypass);
+		proxypass = NULL;
 
 		switch(proxytype) {
 		case PROXY_NONE:
@@ -486,10 +496,11 @@ bool setup_myself_reloadable(void) {
 		case PROXY_EXEC:
 			if(!space || !*space) {
 				logger(DEBUG_ALWAYS, LOG_ERR, "Argument expected for proxy type exec!");
+				free(proxy);
 				return false;
 			}
 
-			proxyhost =  xstrdup(space);
+			proxyhost = xstrdup(space);
 			break;
 
 		case PROXY_SOCKS4:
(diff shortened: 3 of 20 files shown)


hg:
The failing job is the old error which I haven't been able to trigger locally (yet)

    Connected to localhost port 30010...
    Could not connect to ::1 port 30010: Address not available

src/hash.c
```diff
@@ -29,11 +29,11 @@ static uint32_t hash_function(const void *p, size_t len) {
 	uint32_t hash = 0;
 
 	while(true) {
-		for(int i = len > 4 ? 4 : len; --i;) {
+		for(size_t i = len > 4 ? 4 : len; --i;) {
 			hash += (uint32_t)q[len - i] << (8 * i);
 		}
 
-		hash *= 0x9e370001UL; // Golden ratio prime.
+		hash = (uint32_t)(hash * 0x9e370001UL); // Golden ratio prime.
```
gsliepen, line 36: Hm, this should probably just be using the `U` instead of `UL` postfix:

```
hash *= 0x9e370001U;
```

src/logger.c
```diff
@@ -118,6 +117,8 @@ static void real_logger(int level, int priority, const char *message) {
 
 		suppress = false;
 	}
+
+	logcontrol = false;
```
gsliepen, line 121: This looks wrong. The logger function itself should not modify this variable.

src/net_setup.c
```diff
@@ -486,18 +484,30 @@ bool setup_myself_reloadable(void) {
...
-			proxyhost =  xstrdup(space);
+			free(proxyhost);
+			proxyhost = xstrdup(space);
 			break;
 
 		case PROXY_SOCKS4:
 		case PROXY_SOCKS4A:
 		case PROXY_SOCKS5:
 		case PROXY_HTTP:
+			free(proxyhost);
```
gsliepen, line 499: I think all those `free()`s and setting to `NULL` should be done once before the `switch`-statement.

src/net_setup.c
```diff
@@ -512,6 +522,7 @@ bool setup_myself_reloadable(void) {
 
 			if(!proxyhost || !*proxyhost || !proxyport || !*proxyport) {
 				logger(DEBUG_ALWAYS, LOG_ERR, "Host and port argument expected for proxy!");
+				free(proxy);
```
gsliepen, line 525: I think this is also should set `proxyhost` and `proxyport` back to `NULL`.

src/net_setup.c
```diff
@@ -564,12 +580,16 @@ bool setup_myself_reloadable(void) {
 			routing_mode = RMODE_HUB;
 		} else {
 			logger(DEBUG_ALWAYS, LOG_ERR, "Invalid routing mode!");
+			free(rmode);
 			return false;
 		}
 
 		free(rmode);
+		rmode = NULL;
```
gsliepen, line 588: This seems unnecessary, `rmode` is just a local variable that's no longer used after this. If we want to make this more explicit, we could put braces around each block that checks a certain config variable, like so:

```
{
    char *rmode = NULL;

    if(get_config_string(lookup_config(config_tree, "Mode"), &rmode)) {
        ...
    }
}
```

src/utils.c
```diff
@@ -67,7 +67,7 @@ size_t hex2bin(const char *src, void *vdst, size_t length) {
 size_t bin2hex(const void *vsrc, char *dst, size_t length) {
 	const char *src = vsrc;
 
-	for(size_t i = length; i-- > 0;) {
+	for(ssize_t i = (ssize_t) length - 1; i >= 0; --i) {
```
gsliepen, line 70: I don't think this is an improvement. The original code is correct and not UB as far as I can see. Did this trigger a complaint from a sanitizer?

gsliepen (changes requested):
> What are your thoughts on fixing a lot of clang-tidy warnings similar to:
> 
> - Narrowing conversion from 'unsigned long' to signed type 'int' is implementation-defined these in particular often tend to trigger UBSAN at run time

Yes. I think we should use `size_t` consistenly for all things sizes, counts and indices, which should be most of the narrowing conversion issues I guess.

> - 'sscanf' used to convert a string to an integer value, but function will not report conversion errors; consider using 'strtol' instead
> -  'atoi' used to convert a string to an integer value, but function will not report conversion errors; consider using 'strtol' instead

Sure.

> - Unused header_name.h
> - Declaration shadows a local variable

Yes.

> - Declaration shadows a variable in the global scope

This one is harder. It's especially hard to predict what global variables are defined by the standard library on some platforms. If we can easily avoid this issue for tinc's own variables, then sure. But if possible, I'd like to avoid counting this as an error during CI runs.

> - and then adding clang-tidy with -Werror to CI?

Yes, but perhaps not have this immediately cause a fail result of CI runs, maybe add `allow_failure: true` as [suggested here](https://stackoverflow.com/questions/59680450/how-to-detect-compiler-warnings-in-gitlab-ci)? We can always make it a real failure if we are confident we are not triggering this too often, and this will allow us to add this stage without first having to fix everything.

> I would like to reduce the number of these to zero. However, it will produce a large diff which would be difficult to slice into pieces because fixing one warning often requires changing global functions or fields in widely used structs, which produces further warnings, and so on.
> 
> Will you accept patches to clean this up, seeing that this is not absolutely necessary and (as always) has a chance of breaking something?

If we can add the checks without reporting failure, then I'd say add `clang-tidy`, and then we can have pull requests for fixes later, and it doesn't have to be everything in one big blob.

src/utils.c
hg: UBSAN was spamming logs (about 50 reports for each test run):

    SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior net_setup.c:326:17 in
    utils.c:70:26: runtime error: unsigned integer overflow: 0 - 1 cannot be represented in type 'size_t' (aka 'unsigned long')
        #0 0x55b2263c1e88 in bin2hex tinc/src/utils.c:70:26
        #1 0x55b22630a50b in init_control tinc/src/control.c:147:2
        #2 0x55b22634eeba in setup_network tinc/src/net_setup.c:1295:6
        #3 0x55b2263c0456 in main tinc/src/tincd.c:573:6
        #4 0x7f34318690b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)
        #5 0x55b2262d8ded in _start (tinc/src/tincd+0xa9ded)

I'll fix everything you noted, thanks for the review!

gsliepen: Ugh. Unsigned integer overflow is *defined* behavior, at least since C99. But forcibly casting an unsigned to signed to get rid of this warning is not really nice either. Perhaps this can be replaced with:

```
for(size_t i = length; i > 0;) {
    --i;
    ...
}
```

gsliepen: Oh, or should it have been `i - 1U`?

src/hash.c
hg: This I wasn't able to change. If we [replace](https://github.com/hg/tinc/runs/3155144177) this line with

```c
hash *= 0x9e370001U; // Golden ratio prime.
```

or even with an [explicit cast](https://github.com/hg/tinc/runs/3155098787) after the fact

```c
hash = (uint32_t)(hash * 0x9e370001U); // Golden ratio prime.
```

it still does not make the sanitizer happy:

    hash.c:36:8: runtime error: unsigned integer overflow: 512 * 2654404609 cannot be represented in type 'unsigned int'
        #0 0x55ea5adceffe in hash_function src/hash.c:36:8
        #1 0x55ea5adcf166 in hash_search src/hash.c:86:22
        #2 0x55ea5ae6e8b6 in lookup_subnet_ipv6 src/subnet.c:192:10
        ...

It seems to be very strict with overflows and forces us to make them explicit. We can always [weaken the checks](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#silencing-unsigned-integer-overflow), of course.

hg:
Everything else should be fixed. Thanks again for putting in time and patience to review this.

- https://github.com/hg/tinc/actions/runs/1064801500 (reformatted since then)
- https://github.com/hg/tinc/actions/runs/1064812845
- https://builds.sr.ht/~reducer/job/552003
- https://builds.sr.ht/~reducer/job/552004
- https://builds.sr.ht/~reducer/job/552005

---

Another quick question — what is your stance on putting `const` everywhere possible (including non-pointer function arguments and things like `const char *const str`)?

C doesn't make it easy, but it seems to be a worthwhile thing to add as I continue to pick up the low hanging fruit. Or is it just noise?

gsliepen:
> Another quick question — what is your stance on putting const everywhere possible (including non-pointer function arguments and things like const char *const str)?
> 
> C doesn't make it easy, but it seems to be a worthwhile thing to add as I continue to pick up the low hanging fruit. Or is it just noise?

I would give this very low priority. This is most likely not going to find any bugs.

src/hash.c
gsliepen: Since it's just this one one thing, let's keep this check enabled.

PR #296, opened 2021-08-01
Add clang-tidy to CI, fix reported narrowing conversion warnings

hg:
This gets rid of all reported narrowing casts (eliminating them whenever possible, and adding explicit casts to silence the useless warning otherwise), and adjusts variable types where it makes sense to do so:

- `char` → `uint8_t` where memory is expected to contain opaque binary data
- `ssize_t` where it is returned by library functions
- `size_t` for `sizeof(..)` results, storing array indices, etc.
- `ptrdiff_t` for pointer arithmetic results
- and so on

clang-tidy has been added to CI and will fail if it finds anything suspicious. Only narrowing casts are checked for now (see `.clang-tidy`), we can enable more warning types one by one after fixing them.

The full list for you particular version of clang can be obtained with:

    $ clang-tidy --checks='*' --list-checks

or [here](https://clang.llvm.org/extra/clang-tidy/checks/list.html).

You may also find [this page](https://clang.llvm.org/extra/clang-tidy/Integrations.html) useful.

Cryptographic functions are skipped because they produce a ton of warnings and it's probably best not to mess with them.

Solaris and Windows code is also skipped because clang-tidy requires a fully working translation unit with all included headers present. I couldn't build compilation database on Windows (autotools really doesn't make it easy, and most tools to do it don't work there), and we have no CI for Solaris.

When (if) we switch the build system to something more modern that generates the compilation database without third-party tools, this could be updated.

- https://builds.sr.ht/~reducer/job/557195
- https://builds.sr.ht/~reducer/job/557202
- https://builds.sr.ht/~reducer/job/557201
- https://github.com/hg/tinc/actions/runs/1087055817 (with e050753d1fc4d4465032919746c2a4db78f19932 reverted)
- https://github.com/hg/tinc/actions/runs/1087167564
- https://github.com/hg/tinc/actions/runs/1087131224
- https://github.com/hg/tinc/actions/runs/1087042963
- https://github.com/hg/tinc/actions/runs/1086990423
- https://github.com/hg/tinc/actions/runs/1086632473
- https://github.com/hg/tinc/actions/runs/1086801641

These warnings are caused by system headers which we can't do anything about:

```
...
1184 warnings generated.
1201 warnings generated.
Suppressed 1201 warnings (1104 in non-user code, 97 with check filters).
```

For comparison, here's an example of a real issue:

- [FreeBSD](https://builds.sr.ht/~reducer/job/557224#task-lint-125)
- [macOS](https://github.com/hg/tinc/runs/3213173005?check_suite_focus=true#step:6:102)

Linux job won't show this until GitHub update their Ubuntu VMs to clang-tidy 12 or later. It cannot easily be installed manually because of broken dependencies. We can work around this with a separate container, but the best solution IMHO is to simply wait until clang-tidy is updated, and rely on FreeBSD and macOS for now.


src/net_setup.c
```diff
@@ -860,9 +860,9 @@ static bool setup_myself(void) {
 
 	/* Compression */
 
-	if(get_config_int(lookup_config(config_tree, "Compression"), &myself->incompression)) {
+	if(get_config_int(lookup_config(config_tree, "Compression"), (int *) &myself->incompression)) {
```
gsliepen, line 863: While in practice that will work, this is incorrect; the compiler has some freedom in choosing the underlying type of an enum, so it might not be the same size as an `int`. You could use a temporary `int` variable to read it into, and then copy the result into `myself->incompression`.

src/openssl/digest.c
```diff
@@ -27,13 +27,13 @@
 #include "../digest.h"
 #include "../logger.h"
 
-static digest_t *digest_open(const EVP_MD *evp_md, int maclength) {
+static digest_t *digest_open(const EVP_MD *evp_md, ssize_t maclength) {
```
gsliepen, line 30: Changing `maclength` to `ssize_t` doesn't really improve things, it still needs a lot of casting. Perhaps it is better to make it `size_t`. While we check for it to be negative, nowhere do we actually set it to a negative value, so I don't think we need to handle negative values at all.

src/protocol.h
```diff
@@ -107,7 +107,7 @@ extern void send_key_changed(void);
 extern bool send_req_key(struct node_t *to);
 extern bool send_ans_key(struct node_t *to);
 extern bool send_tcppacket(struct connection_t *c, const struct vpn_packet_t *packet);
-extern bool send_sptps_tcppacket(struct connection_t *c, const char *packet, int len);
+extern bool send_sptps_tcppacket(struct connection_t *c, const char *packet, size_t len);
```
gsliepen, line 110: Maybe the `const char *` should become a `const void *` or `const uint8_t *` as well?

src/tincd.c
```diff
@@ -506,7 +508,7 @@ int main(int argc, char **argv) {
 	}
 
 	if(!debug_level) {
-		get_config_int(lookup_config(config_tree, "LogLevel"), &debug_level);
+		get_config_int(lookup_config(config_tree, "LogLevel"), (int *) &debug_level);
```
gsliepen, line 511: Same issue with `enum`s not necessarily having the same size as `int` here.

src/utils.c
```diff
@@ -45,11 +45,11 @@ static const char base64_decode[256] = {
 	        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
         };
 
-static int charhex2bin(char c) {
+static uint8_t charhex2bin(char c) {
 	if(isdigit(c)) {
 		return c - '0';
 	} else {
-		return toupper(c) - 'A' + 10;
+		return toupper((uint8_t) c) - 'A' + 10;
```
gsliepen, line 52: Ah yes, the dreaded ctype.h functions. Casting it to `uint8_t` is indeed correct, it should also be done for `isdigit()` in line 49 though, and there are some other `is*()` calls in the code that need fixing.

src/vde_device.c
```diff
@@ -102,7 +101,7 @@ static void close_device(void) {
 }
 
 static bool read_packet(vpn_packet_t *packet) {
-	int lenin = (ssize_t)plug.vde_recv(conn, DATA(packet), MTU, 0);
+	ssize_t lenin = (ssize_t) plug.vde_recv(conn, DATA(packet), MTU, 0);
```
gsliepen, line 104: Hm, I don't know why the explicit cast was necessary... according to the manpage, `vde_recv()` already returns a `ssize_t`...

gsliepen (changes requested):

src/vde_device.c
hg: This was triggered by testing on one of the supported platforms (probably FreeBSD). The latest version does indeed declare the returned type as `ssize_t`:

https://github.com/bringhurst/vde/blob/master/include/libvdeplug_dyn.h#L89

However, the same header on my desktop OS has `size_t` there. FreeBSD 13 has `size_t`, too.

They wiped their change history when moving from svn to git, so here's that change in someone else's fork of the project:

https://github.com/bringhurst/vde/commit/385d4b49b05c4b8b2ecdc695ec42a863024e7da6#diff-8744b91559c96b744dd3f015a4cd367daccb11a92219e9e4f4cb2d501cf36a30

src/openssl/digest.c
hg: `digest_open` is called by `digest_open_by_name` and `digest_open_by_nid`, both of which are called with `maclength == -1`.

https://github.com/gsliepen/tinc/blob/1.1/src/openssl/prf.c#L33

https://github.com/gsliepen/tinc/blob/1.1/src/protocol_auth.c#L642

https://github.com/gsliepen/tinc/blob/1.1/src/protocol_auth.c#L527

Maybe it's best to replace `maclength` with `size_t` and change all these calls to `0`?

hg:
Huge thanks for taking the time to review this. I fixed everything except for [this](https://github.com/gsliepen/tinc/pull/296#discussion_r680516935) (I'll await your judgement on that), plus all calls to functions from `ctype` and unnecessary casts to unsigned int (instead of `%zu`).

Diff:

```diff
diff --git a/src/Makefile.am b/src/Makefile.am
index 306bbc42..5165b583 100644
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -52,6 +52,7 @@ tincd_SOURCES = \
 	autoconnect.c autoconnect.h \
 	buffer.c buffer.h \
 	cipher.h \
+	compression.h \
 	conf.c conf.h \
 	conf_net.c conf_net.h \
 	connection.c connection.h \
diff --git a/src/autoconnect.c b/src/autoconnect.c
index ae514286..d819ee46 100644
--- a/src/autoconnect.c
+++ b/src/autoconnect.c
@@ -29,7 +29,7 @@ static void make_new_connection() {
 	int count = 0;
 
 	for splay_each(node_t, n, node_tree) {
-		if(n == myself || n->connection || !(n->status.has_address != 0u || n->status.reachable != 0u)) {
+		if(n == myself || n->connection || !(n->status.has_address || n->status.reachable)) {
 			continue;
 		}
 
@@ -43,7 +43,7 @@ static void make_new_connection() {
 	int r = rand() % count;
 
 	for splay_each(node_t, n, node_tree) {
-		if(n == myself || n->connection || !(n->status.has_address != 0u || n->status.reachable != 0u)) {
+		if(n == myself || n->connection || !(n->status.has_address || n->status.reachable)) {
 			continue;
 		}
 
@@ -88,7 +88,7 @@ static void connect_to_unreachable() {
 		}
 
 		/* Is it unreachable and do we know an address for it? If not, return. */
-		if(n == myself || n->connection || n->status.reachable != 0u || !n->status.has_address) {
+		if(n == myself || n->connection || n->status.reachable || !n->status.has_address) {
 			return;
 		}
 
diff --git a/src/bsd/tunemu.c b/src/bsd/tunemu.c
index 7d1c407a..6867c47b 100644
--- a/src/bsd/tunemu.c
+++ b/src/bsd/tunemu.c
@@ -50,10 +50,10 @@
 #define PPPIOCGUNIT     _IOR('t', 86, int)
 
 struct sockaddr_ppp {
-	u_int8_t ppp_len;
-	u_int8_t ppp_family;
-	u_int16_t ppp_proto;
-	u_int32_t ppp_cookie;
+	uint8_t ppp_len;
+	uint8_t ppp_family;
+	uint16_t ppp_proto;
+	uint32_t ppp_cookie;
 };
 
 enum NPmode {
diff --git a/src/compression.h b/src/compression.h
new file mode 100644
index 00000000..96fc9ea8
--- /dev/null
+++ b/src/compression.h
@@ -0,0 +1,25 @@
+#ifndef TINC_COMPRESSION_H
+#define TINC_COMPRESSION_H
+
+typedef enum compression_level_t {
+	COMPRESS_NONE = 0,
+
+	COMPRESS_ZLIB_1 = 1,
+	COMPRESS_ZLIB_2 = 2,
+	COMPRESS_ZLIB_3 = 3,
+	COMPRESS_ZLIB_4 = 4,
+	COMPRESS_ZLIB_5 = 5,
+	COMPRESS_ZLIB_6 = 6,
+	COMPRESS_ZLIB_7 = 7,
+	COMPRESS_ZLIB_8 = 8,
+	COMPRESS_ZLIB_9 = 9,
+
+	COMPRESS_LZO_LO = 10,
+	COMPRESS_LZO_HI = 11,
+
+	COMPRESS_LZ4 = 12,
+
+	COMPRESS_GUARD = INT_MAX, /* ensure that sizeof(compression_level_t) == sizeof(int) */
+} compression_level_t;
+
+#endif
diff --git a/src/connection.h b/src/connection.h
index 506e252f..2f652182 100644
--- a/src/connection.h
+++ b/src/connection.h
@@ -27,6 +27,7 @@
 #include "rsa.h"
 #include "list.h"
 #include "sptps.h"
+#include "compression.h"
 
 #define OPTION_INDIRECT         0x0001
 #define OPTION_TCPONLY          0x0002
@@ -35,21 +36,21 @@
 #define OPTION_VERSION(x) ((x) >> 24) /* Top 8 bits are for protocol minor version */
 
 typedef struct connection_status_t {
-	uint32_t pinged: 1;                 /* sent ping */
-	uint32_t unused_active: 1;
-	uint32_t connecting: 1;             /* 1 if we are waiting for a non-blocking connect() to finish */
-	uint32_t unused_termreq: 1;         /* the termination of this connection was requested */
-	uint32_t remove_unused: 1;          /* Set to 1 if you want this connection removed */
-	uint32_t timeout_unused: 1;         /* 1 if gotten timeout */
-	uint32_t encryptout: 1;             /* 1 if we can encrypt outgoing traffic */
-	uint32_t decryptin: 1;              /* 1 if we have to decrypt incoming traffic */
-	uint32_t mst: 1;                    /* 1 if this connection is part of a minimum spanning tree */
-	uint32_t control: 1;                /* 1 if this is a control connection */
-	uint32_t pcap: 1;                   /* 1 if this is a control connection requesting packet capture */
-	uint32_t log: 1;                    /* 1 if this is a control connection requesting log dump */
-	uint32_t invitation: 1;             /* 1 if this is an invitation */
-	uint32_t invitation_used: 1;        /* 1 if the invitation has been consumed */
-	uint32_t tarpit: 1;                 /* 1 if the connection should be added to the tarpit */
+	bool pinged: 1;                 /* sent ping */
+	bool unused_active: 1;
+	bool connecting: 1;             /* 1 if we are waiting for a non-blocking connect() to finish */
+	bool unused_termreq: 1;         /* the termination of this connection was requested */
+	bool remove_unused: 1;          /* Set to 1 if you want this connection removed */
+	bool timeout_unused: 1;         /* 1 if gotten timeout */
+	bool encryptout: 1;             /* 1 if we can encrypt outgoing traffic */
+	bool decryptin: 1;              /* 1 if we have to decrypt incoming traffic */
+	bool mst: 1;                    /* 1 if this connection is part of a minimum spanning tree */
+	bool control: 1;                /* 1 if this is a control connection */
+	bool pcap: 1;                   /* 1 if this is a control connection requesting packet capture */
+	bool log: 1;                    /* 1 if this is a control connection requesting log dump */
+	bool invitation: 1;             /* 1 if this is an invitation */
+	bool invitation_used: 1;        /* 1 if the invitation has been consumed */
+	bool tarpit: 1;                 /* 1 if the connection should be added to the tarpit */
 	uint32_t unused: 17;
 } connection_status_t;
 
@@ -90,7 +91,7 @@ typedef struct connection_t {
 	sptps_t sptps;
 
 	int outmaclength;
-	int outcompression;
+	compression_level_t outcompression;
 
 	uint8_t *hischallenge;             /* The challenge we sent to him */
 	uint8_t *mychallenge;              /* The challenge we received */
diff --git a/src/fd_device.c b/src/fd_device.c
index 0380f45f..7360d17e 100644
--- a/src/fd_device.c
+++ b/src/fd_device.c
@@ -83,8 +83,8 @@ static int read_fd(int socket) {
 	}
 
 	if(cmsgptr->cmsg_len != CMSG_LEN(sizeof(device_fd))) {
-		logger(DEBUG_ALWAYS, LOG_ERR, "Wrong CMSG data length: %lu, expected %lu!",
-		       (unsigned long)cmsgptr->cmsg_len, (unsigned long)CMSG_LEN(sizeof(device_fd)));
+		logger(DEBUG_ALWAYS, LOG_ERR, "Wrong CMSG data length: %zu, expected %zu!",
+		       cmsgptr->cmsg_len, CMSG_LEN(sizeof(device_fd)));
 		return -1;
 	}
 
diff --git a/src/gcrypt/digest.c b/src/gcrypt/digest.c
index f790c8fb..383842f6 100644
--- a/src/gcrypt/digest.c
+++ b/src/gcrypt/digest.c
@@ -143,7 +143,7 @@ bool digest_create(digest_t *digest, const void *indata, size_t inlen, void *out
 	unsigned int len = gcry_md_get_algo_dlen(digest->algo);
 
 	if(digest->hmac) {
-		u_int8_t *tmpdata;
+		uint8_t *tmpdata;
 		gcry_md_reset(digest->hmac);
 		gcry_md_write(digest->hmac, indata, inlen);
 		tmpdata = gcry_md_read(digest->hmac, digest->algo);
@@ -164,7 +164,7 @@ bool digest_create(digest_t *digest, const void *indata, size_t inlen, void *out
 
 bool digest_verify(digest_t *digest, const void *indata, size_t inlen, const void *cmpdata) {
 	ssize_t len = digest->maclength;
-	u_int8_t outdata[len];
+	uint8_t outdata[len];
 
 	return digest_create(digest, indata, inlen, outdata) && !memcmp(cmpdata, outdata, len);
 }
diff --git a/src/gcrypt/rsa.c b/src/gcrypt/rsa.c
index e44b3438..c026968a 100644
--- a/src/gcrypt/rsa.c
+++ b/src/gcrypt/rsa.c
@@ -297,7 +297,8 @@ bool rsa_public_encrypt(rsa_t *rsa, void *in, size_t len, void *out) {
 	gcry_mpi_t outmpi = gcry_mpi_new(len * 8);
 	gcry_mpi_powm(outmpi, inmpi, rsa->e, rsa->n);
 
-	size_t pad = len - (gcry_mpi_get_nbits(outmpi) + 7) / 8;
+	size_t out_bytes = (gcry_mpi_get_nbits(outmpi) + 7) / 8;
+	size_t pad = len - MIN(out_bytes, len);
 
 	while(pad--) {
 		*(char *)out++ = 0;
diff --git a/src/graph.c b/src/graph.c
index edd9fc78..4de9181c 100644
--- a/src/graph.c
+++ b/src/graph.c
@@ -90,7 +90,7 @@ static void mst_kruskal(void) {
 	bool skipped = false;
 
 	for splay_each(edge_t, e, edge_weight_tree) {
-		if(!e->reverse || ((uint32_t) e->from->status.visited == e->to->status.visited)) {
+		if(!e->reverse || (e->from->status.visited == e->to->status.visited)) {
 			skipped = true;
 			continue;
 		}
@@ -171,9 +171,9 @@ static void sssp_bfs(void) {
 			     of nodes behind it.
 			 */
 
-			bool indirect = n->status.indirect != 0u || e->options & OPTION_INDIRECT;
+			bool indirect = n->status.indirect || e->options & OPTION_INDIRECT;
 
-			if(e->to->status.visited != 0u
+			if(e->to->status.visited
 			                && (!e->to->status.indirect || indirect)
 			                && (e->to->distance != n->distance + 1 || e->weight >= e->to->prevedge->weight)) {
 				continue;
@@ -214,7 +214,7 @@ static void check_reachability(void) {
 	int became_unreachable_count = 0;
 
 	for splay_each(node_t, n, node_tree) {
-		if((int) n->status.visited != (int) n->status.reachable) {
+		if(n->status.visited != n->status.reachable) {
 			n->status.reachable = !n->status.reachable;
 			n->last_state_change = now.tv_sec;
 
@@ -268,9 +268,9 @@ static void check_reachability(void) {
 			environment_add(&env, "REMOTEADDRESS=%s", address);
 			environment_add(&env, "REMOTEPORT=%s", port);
 
-			execute_script(n->status.reachable != 0u ? "host-up" : "host-down", &env);
+			execute_script(n->status.reachable ? "host-up" : "host-down", &env);
 
-			xasprintf(&name, n->status.reachable != 0u ? "hosts/%s-up" : "hosts/%s-down", n->name);
+			xasprintf(&name, n->status.reachable ? "hosts/%s-up" : "hosts/%s-down", n->name);
 			execute_script(name, &env);
 
 			free(name);
@@ -292,7 +292,7 @@ static void check_reachability(void) {
 			}
 		}
 
-		if(n->status.reachable != 0u && n != myself) {
+		if(n->status.reachable && n != myself) {
 			reachable_count++;
 		}
 	}
diff --git a/src/invitation.c b/src/invitation.c
index e2384013..3c23a170 100644
--- a/src/invitation.c
+++ b/src/invitation.c
@@ -170,7 +170,7 @@ char *get_my_hostname() {
 	// Check that the hostname is reasonable
 	if(hostname) {
 		for(char *p = hostname; *p; p++) {
-			if(isalnum(*p) || *p == '-' || *p == '.' || *p == ':') {
+			if(isalnum((uint8_t) *p) || *p == '-' || *p == '.' || *p == ':') {
 				continue;
 			}
 
@@ -216,7 +216,7 @@ again:
 	}
 
 	for(char *p = line; *p; p++) {
-		if(isalnum(*p) || *p == '-' || *p == '.') {
+		if(isalnum((uint8_t) *p) || *p == '-' || *p == '.') {
 			continue;
 		}
 
@@ -572,7 +572,7 @@ static char *get_line(const char **data) {
 		return NULL;
 	}
 
-	if(len && !isprint(**data)) {
+	if(len && !isprint((uint8_t) **data)) {
 		abort();
 	}
 
diff --git a/src/logger.c b/src/logger.c
index 9f9e3f2d..587022a7 100644
--- a/src/logger.c
+++ b/src/logger.c
@@ -106,7 +106,7 @@ static void real_logger(debug_t level, int priority, const char *message) {
 
 			logcontrol = true;
 
-			if(level > (c->outcompression >= 0 ? c->outcompression : debug_level)) {
+			if(level > (c->outcompression >= COMPRESS_NONE ? c->outcompression : debug_level)) {
 				continue;
 			}
 
diff --git a/src/meta.c b/src/meta.c
index debeab45..afc98ab2 100644
--- a/src/meta.c
+++ b/src/meta.c
@@ -56,8 +56,8 @@ bool send_meta(connection_t *c, const void *buffer, size_t length) {
 		abort();
 	}
 
-	logger(DEBUG_META, LOG_DEBUG, "Sending %lu bytes of metadata to %s (%s)", (unsigned long)length,
-	       c->name, c->hostname);
+	logger(DEBUG_META, LOG_DEBUG, "Sending %zu bytes of metadata to %s (%s)",
+	       length, c->name, c->hostname);
 
 	if(c->protocol_minor >= 2) {
 		return sptps_send_record(&c->sptps, 0, buffer, length);
@@ -100,8 +100,8 @@ void send_meta_raw(connection_t *c, const void *buffer, size_t length) {
 		abort();
 	}
 
-	logger(DEBUG_META, LOG_DEBUG, "Sending %lu bytes of raw metadata to %s (%s)", (unsigned long)length,
-	       c->name, c->hostname);
+	logger(DEBUG_META, LOG_DEBUG, "Sending %zu bytes of raw metadata to %s (%s)",
+	       length, c->name, c->hostname);
 
 	buffer_add(&c->outbuf, buffer, length);
 
diff --git a/src/net_packet.c b/src/net_packet.c
index 3998dbeb..dab74e5d 100644
--- a/src/net_packet.c
+++ b/src/net_packet.c
@@ -41,6 +41,7 @@
 #include "cipher.h"
 #include "conf.h"
 #include "connection.h"
+#include "compression.h"
 #include "crypto.h"
 #include "digest.h"
 #include "device.h"
@@ -955,8 +956,8 @@ bool send_sptps_data(node_t *to, node_t *from, int type, const void *data, size_
 
 	if(type == SPTPS_HANDSHAKE || tcponly || (!direct && !relay_supported) || (type != PKT_PROBE && (len - SPTPS_DATAGRAM_OVERHEAD) > relay->minmtu)) {
 		if(type != SPTPS_HANDSHAKE && (to->nexthop->connection->options >> 24) >= 7) {
-			char buf[len + sizeof(to->id) + sizeof(from->id)];
-			char *buf_ptr = buf;
+			uint8_t buf[len + sizeof(to->id) + sizeof(from->id)];
+			uint8_t *buf_ptr = buf;
 			memcpy(buf_ptr, &to->id, sizeof(to->id));
 			buf_ptr += sizeof(to->id);
 			memcpy(buf_ptr, &from->id, sizeof(from->id));
@@ -1137,7 +1138,7 @@ bool receive_sptps_record(void *handle, uint8_t type, const void *data, uint16_t
 		}
 	}
 
-	if(from->status.udppacket != 0u && inpkt.len > from->maxrecentlen) {
+	if(from->status.udppacket && inpkt.len > from->maxrecentlen) {
 		from->maxrecentlen = inpkt.len;
 	}
 
@@ -1195,7 +1196,7 @@ static void try_udp(node_t *n) {
 
 	/* Send gratuitous probe replies to 1.1 nodes. */
 
-	if((n->options >> 24) >= 3 && n->status.udp_confirmed != 0u) {
+	if((n->options >> 24) >= 3 && n->status.udp_confirmed) {
 		struct timeval ping_tx_elapsed;
 		timersub(&now, &n->udp_reply_sent, &ping_tx_elapsed);
 
@@ -1219,7 +1220,7 @@ static void try_udp(node_t *n) {
 	struct timeval ping_tx_elapsed;
 	timersub(&now, &n->udp_ping_sent, &ping_tx_elapsed);
 
-	int interval = n->status.udp_confirmed != 0u
+	int interval = n->status.udp_confirmed
 	               ? udp_discovery_keepalive_interval
 	               : udp_discovery_interval;
 
@@ -1623,7 +1624,7 @@ void broadcast_packet(const node_t *from, vpn_packet_t *packet) {
 	// usually distributes the sending of broadcast packets over all nodes.
 	case BMODE_MST:
 		for list_each(connection_t, c, connection_list)
-			if(c->edge && c->status.mst != 0u && c != from->nexthop->connection) {
+			if(c->edge && c->status.mst && c != from->nexthop->connection) {
 				send_packet(c->node, packet);
 			}
 
@@ -1638,7 +1639,7 @@ void broadcast_packet(const node_t *from, vpn_packet_t *packet) {
 		}
 
 		for splay_each(node_t, n, node_tree)
-			if(n->status.reachable != 0u && n != myself && ((n->via == myself && n->nexthop == n) || n->via == n)) {
+			if(n->status.reachable && n != myself && ((n->via == myself && n->nexthop == n) || n->via == n)) {
 				send_packet(n, packet);
 			}
 
@@ -1665,7 +1666,7 @@ static node_t *try_harder(const sockaddr_t *from, const vpn_packet_t *pkt) {
 			continue;
 		}
 
-		if(!n->status.validkey_in && !(n->status.sptps != 0u && n->sptps.instate)) {
+		if(!n->status.validkey_in && !(n->status.sptps && n->sptps.instate)) {
 			continue;
 		}
 
@@ -1726,7 +1727,7 @@ static void handle_incoming_vpn_packet(listen_socket_t *ls, vpn_packet_t *pkt, s
 		pkt->offset = 2 * sizeof(node_id_t);
 		from = lookup_node_id(SRCID(pkt));
 
-		if(from && from->status.sptps != 0u && !memcmp(DSTID(pkt), &nullid, sizeof(nullid))) {
+		if(from && from->status.sptps && !memcmp(DSTID(pkt), &nullid, sizeof(nullid))) {
 			if(sptps_verify_datagram(&from->sptps, DATA(pkt), pkt->len - 2 * sizeof(node_id_t))) {
 				n = from;
 			} else {
diff --git a/src/net_setup.c b/src/net_setup.c
index 42a67a7f..90a78495 100644
--- a/src/net_setup.c
+++ b/src/net_setup.c
@@ -26,6 +26,7 @@
 #include "conf_net.h"
 #include "conf.h"
 #include "connection.h"
+#include "compression.h"
 #include "control.h"
 #include "device.h"
 #include "digest.h"
@@ -859,8 +860,9 @@ static bool setup_myself(void) {
 #endif
 
 	/* Compression */
+	int compression = COMPRESS_NONE;
 
-	if(get_config_int(lookup_config(config_tree, "Compression"), (int *) &myself->incompression)) {
+	if(get_config_int(lookup_config(config_tree, "Compression"), &compression)) {
 		switch(myself->incompression) {
 		case COMPRESS_LZ4:
 #ifdef HAVE_LZ4
@@ -906,6 +908,8 @@ static bool setup_myself(void) {
 			logger(DEBUG_ALWAYS, LOG_ERR, "Compression level %i is unrecognized by this node.", myself->incompression);
 			return false;
 		}
+
+		myself->incompression = compression;
 	} else {
 		myself->incompression = COMPRESS_NONE;
 	}
diff --git a/src/node.h b/src/node.h
index f544ea09..f990c326 100644
--- a/src/node.h
+++ b/src/node.h
@@ -27,42 +27,22 @@
 #include "digest.h"
 #include "event.h"
 #include "subnet.h"
-
-typedef enum compression_level_t {
-	COMPRESS_NONE = 0,
-
-	COMPRESS_ZLIB_1 = 1,
-	COMPRESS_ZLIB_2 = 2,
-	COMPRESS_ZLIB_3 = 3,
-	COMPRESS_ZLIB_4 = 4,
-	COMPRESS_ZLIB_5 = 5,
-	COMPRESS_ZLIB_6 = 6,
-	COMPRESS_ZLIB_7 = 7,
-	COMPRESS_ZLIB_8 = 8,
-	COMPRESS_ZLIB_9 = 9,
-
-	COMPRESS_LZO_LO = 10,
-	COMPRESS_LZO_HI = 11,
-
-	COMPRESS_LZ4 = 12,
-
-	COMPRESS_GUARD = INT_MAX, /* ensure that sizeof(compression_level_t) == sizeof(int) */
-} compression_level_t;
+#include "compression.h"
 
 typedef struct node_status_t {
-	uint32_t unused_active: 1;          /* 1 if active (not used for nodes) */
-	uint32_t validkey: 1;               /* 1 if we currently have a valid key for him */
-	uint32_t waitingforkey: 1;          /* 1 if we already sent out a request */
-	uint32_t visited: 1;                /* 1 if this node has been visited by one of the graph algorithms */
-	uint32_t reachable: 1;              /* 1 if this node is reachable in the graph */
-	uint32_t indirect: 1;               /* 1 if this node is not directly reachable by us */
-	uint32_t sptps: 1;                  /* 1 if this node supports SPTPS */
-	uint32_t udp_confirmed: 1;          /* 1 if the address is one that we received UDP traffic on */
-	uint32_t send_locally: 1;           /* 1 if the next UDP packet should be sent on the local network */
-	uint32_t udppacket: 1;              /* 1 if the most recently received packet was UDP */
-	uint32_t validkey_in: 1;            /* 1 if we have sent a valid key to him */
-	uint32_t has_address: 1;            /* 1 if we know an external address for this node */
-	uint32_t ping_sent: 1;              /* 1 if we sent a UDP probe but haven't received the reply yet */
+	bool unused_active: 1;          /* 1 if active (not used for nodes) */
+	bool validkey: 1;               /* 1 if we currently have a valid key for him */
+	bool waitingforkey: 1;          /* 1 if we already sent out a request */
+	bool visited: 1;                /* 1 if this node has been visited by one of the graph algorithms */
+	bool reachable: 1;              /* 1 if this node is reachable in the graph */
+	bool indirect: 1;               /* 1 if this node is not directly reachable by us */
+	bool sptps: 1;                  /* 1 if this node supports SPTPS */
+	bool udp_confirmed: 1;          /* 1 if the address is one that we received UDP traffic on */
+	bool send_locally: 1;           /* 1 if the next UDP packet should be sent on the local network */
+	bool udppacket: 1;              /* 1 if the most recently received packet was UDP */
+	bool validkey_in: 1;            /* 1 if we have sent a valid key to him */
+	bool has_address: 1;            /* 1 if we know an external address for this node */
+	bool ping_sent: 1;              /* 1 if we sent a UDP probe but haven't received the reply yet */
 	uint32_t unused: 19;
 } node_status_t;
 
diff --git a/src/protocol.h b/src/protocol.h
index 62bfe1a8..4ea0c8c7 100644
--- a/src/protocol.h
+++ b/src/protocol.h
@@ -107,7 +107,7 @@ extern void send_key_changed(void);
 extern bool send_req_key(struct node_t *to);
 extern bool send_ans_key(struct node_t *to);
 extern bool send_tcppacket(struct connection_t *c, const struct vpn_packet_t *packet);
-extern bool send_sptps_tcppacket(struct connection_t *c, const char *packet, size_t len);
+extern bool send_sptps_tcppacket(struct connection_t *c, const void *packet, size_t len);
 extern bool send_udp_info(struct node_t *from, struct node_t *to);
 extern bool send_mtu_info(struct node_t *from, struct node_t *to, int mtu);
 
diff --git a/src/protocol_auth.c b/src/protocol_auth.c
index dc87f54d..c5ba071d 100644
--- a/src/protocol_auth.c
+++ b/src/protocol_auth.c
@@ -231,11 +231,11 @@ static bool receive_invitation_sptps(void *handle, uint8_t type, const void *dat
 		return true;
 	}
 
-	if(type == 1 && c->status.invitation_used != 0u) {
+	if(type == 1 && c->status.invitation_used) {
 		return finalize_invitation(c, data, len);
 	}
 
-	if(type != 0 || len != 18 || c->status.invitation_used != 0u) {
+	if(type != 0 || len != 18 || c->status.invitation_used) {
 		return false;
 	}
 
diff --git a/src/protocol_key.c b/src/protocol_key.c
index 223a1213..65fd35ec 100644
--- a/src/protocol_key.c
+++ b/src/protocol_key.c
@@ -39,7 +39,7 @@ void send_key_changed(void) {
 	/* Immediately send new keys to directly connected nodes to keep UDP mappings alive */
 
 	for list_each(connection_t, c, connection_list) {
-		if(c->edge && c->node && c->node->status.reachable != 0u && !c->node->status.sptps) {
+		if(c->edge && c->node && c->node->status.reachable && !c->node->status.sptps) {
 			send_ans_key(c->node);
 		}
 	}
@@ -50,7 +50,7 @@ void send_key_changed(void) {
 
 	if(experimental) {
 		for splay_each(node_t, n, node_tree) {
-			if(n->status.reachable != 0u && n->status.validkey != 0u && n->status.sptps != 0u) {
+			if(n->status.reachable && n->status.validkey && n->status.sptps) {
 				sptps_force_kex(&n->sptps);
 			}
 		}
diff --git a/src/protocol_misc.c b/src/protocol_misc.c
index 5c957533..7994f00d 100644
--- a/src/protocol_misc.c
+++ b/src/protocol_misc.c
@@ -110,7 +110,7 @@ bool tcppacket_h(connection_t *c, const char *request) {
 	return true;
 }
 
-bool send_sptps_tcppacket(connection_t *c, const char *packet, size_t len) {
+bool send_sptps_tcppacket(connection_t *c, const void *packet, size_t len) {
 	/* If there already is a lot of data in the outbuf buffer, discard this packet.
 	   We use a very simple Random Early Drop algorithm. */
 
diff --git a/src/solaris/device.c b/src/solaris/device.c
index f27954b1..d1e114ae 100644
--- a/src/solaris/device.c
+++ b/src/solaris/device.c
@@ -104,7 +104,7 @@ static bool setup_device(void) {
 	char *ptr = device;
 	get_config_string(lookup_config(config_tree, "Interface"), &ptr);
 
-	while(*ptr && !isdigit(*ptr)) {
+	while(*ptr && !isdigit((uint8_t) *ptr)) {
 		ptr++;
 	}
 
diff --git a/src/subnet.c b/src/subnet.c
index 5e4c1b0e..13c46030 100644
--- a/src/subnet.c
+++ b/src/subnet.c
@@ -132,7 +132,7 @@ subnet_t *lookup_subnet_mac(const node_t *owner, const mac_t *address) {
 		if(!memcmp(address, &p->net.mac.address, sizeof(*address))) {
 			r = p;
 
-			if(!p->owner || p->owner->status.reachable != 0u) {
+			if(!p->owner || p->owner->status.reachable) {
 				break;
 			}
 		}
@@ -166,7 +166,7 @@ subnet_t *lookup_subnet_ipv4(const ipv4_t *address) {
 		if(!maskcmp(address, &p->net.ipv4.address, p->net.ipv4.prefixlength)) {
 			r = p;
 
-			if(!p->owner || p->owner->status.reachable != 0u) {
+			if(!p->owner || p->owner->status.reachable) {
 				break;
 			}
 		}
@@ -200,7 +200,7 @@ subnet_t *lookup_subnet_ipv6(const ipv6_t *address) {
 		if(!maskcmp(address, &p->net.ipv6.address, p->net.ipv6.prefixlength)) {
 			r = p;
 
-			if(!p->owner || p->owner->status.reachable != 0u) {
+			if(!p->owner || p->owner->status.reachable) {
 				break;
 			}
 		}
diff --git a/src/tincd.c b/src/tincd.c
index 3355067c..95872c34 100644
--- a/src/tincd.c
+++ b/src/tincd.c
@@ -507,8 +507,12 @@ int main(int argc, char **argv) {
 		return 1;
 	}
 
-	if(!debug_level) {
-		get_config_int(lookup_config(config_tree, "LogLevel"), (int *) &debug_level);
+	if(debug_level == DEBUG_NOTHING) {
+		int level = 0;
+
+		if(get_config_int(lookup_config(config_tree, "LogLevel"), &level)) {
+			debug_level = level;
+		}
 	}
 
 #ifdef HAVE_LZO
diff --git a/src/utils.c b/src/utils.c
index 71359f23..57983241 100644
--- a/src/utils.c
+++ b/src/utils.c
@@ -46,10 +46,12 @@ static const char base64_decode[256] = {
         };
 
 static uint8_t charhex2bin(char c) {
-	if(isdigit(c)) {
-		return c - '0';
+	uint8_t cu = (uint8_t) c;
+
+	if(isdigit(cu)) {
+		return cu - '0';
 	} else {
-		return toupper((uint8_t) c) - 'A' + 10;
+		return toupper(cu) - 'A' + 10;
 	}
 }
 
@@ -57,7 +59,7 @@ size_t hex2bin(const char *src, void *vdst, size_t length) {
 	uint8_t *dst = vdst;
 	size_t i;
 
-	for(i = 0; i < length && isxdigit(src[i * 2]) && isxdigit(src[i * 2 + 1]); i++) {
+	for(i = 0; i < length && isxdigit((uint8_t) src[i * 2]) && isxdigit((uint8_t) src[i * 2 + 1]); i++) {
 		dst[i] = charhex2bin(src[i * 2]) * 16 + charhex2bin(src[i * 2 + 1]);
 	}
 
@@ -210,7 +212,7 @@ bool check_id(const char *id) {
 	}
 
 	for(; *id; id++)
-		if(!isalnum(*id) && *id != '_') {
+		if(!isalnum((uint8_t) *id) && *id != '_') {
 			return false;
 		}
 
@@ -223,7 +225,7 @@ bool check_netname(const char *netname, bool strict) {
 	}
 
 	for(const char *c = netname; *c; c++) {
-		if(iscntrl(*c)) {
+		if(iscntrl((uint8_t) *c)) {
 			return false;
 		}
 
@@ -269,7 +271,7 @@ char *replace_name(const char *name) {
 		ret_name = xstrdup(envname);
 
 		for(char *c = ret_name; *c; c++)
-			if(!isalnum(*c)) {
+			if(!isalnum((uint8_t) *c)) {
 				*c = '_';
 			}
 	} else {

```

src/openssl/digest.c
gsliepen: Ah, I missed that. -1 means using the same size as the digest algorithm produces, 0 means not using a digest. A large value should work as well, like `(size_t)-1`. Maybe a `#define` or `static const` variable could be made for this.

gsliepen:
> Huge thanks for taking the time to review this. I fixed everything except for this (I'll await your judgement on that), plus all calls to functions from ctype and unnecessary casts to unsigned int (instead of `%zu`).

No thank you for putting in the time to do all these cleanups, it is very much appreciated!

hg:
I think this covers everything for now.

```diff
diff --git a/src/digest.h b/src/digest.h
index c7ffdf0b..1cebf582 100644
--- a/src/digest.h
+++ b/src/digest.h
@@ -23,13 +23,14 @@
 #include "system.h"
 
 #define DIGEST_MAX_SIZE 64
+#define DIGEST_ALGO_SIZE ((size_t) -1)
 
 #ifndef DISABLE_LEGACY
 
 typedef struct digest digest_t;
 
-extern digest_t *digest_open_by_name(const char *name, int maclength) __attribute__((__malloc__));
-extern digest_t *digest_open_by_nid(int nid, ssize_t maclength) __attribute__((__malloc__));
+extern digest_t *digest_open_by_name(const char *name, size_t maclength) __attribute__((__malloc__));
+extern digest_t *digest_open_by_nid(int nid, size_t maclength) __attribute__((__malloc__));
 extern void digest_close(digest_t *digest);
 extern bool digest_create(digest_t *digest, const void *indata, size_t inlen, void *outdata) __attribute__((__warn_unused_result__));
 extern bool digest_verify(digest_t *digest, const void *indata, size_t inlen, const void *digestdata) __attribute__((__warn_unused_result__));
@@ -39,6 +40,6 @@ extern size_t digest_keylength(const digest_t *digest);
 extern size_t digest_length(const digest_t *digest);
 extern bool digest_active(const digest_t *digest);
 
-#endif
+#endif // DISABLE_LEGACY
 
-#endif
+#endif // TINC_DIGEST_H
diff --git a/src/gcrypt/digest.c b/src/gcrypt/digest.c
index 383842f6..3333f5f2 100644
--- a/src/gcrypt/digest.c
+++ b/src/gcrypt/digest.c
@@ -24,7 +24,7 @@
 
 static struct {
 	const char *name;
-	int algo;
+	enum gcry_md_algos algo;
 	int nid;
 } digesttable[] = {
 	{"none", GCRY_MD_NONE, 0},
@@ -34,7 +34,7 @@ static struct {
 	{"sha512", GCRY_MD_SHA512, 674},
 };
 
-static bool nametodigest(const char *name, int *algo) {
+static bool nametodigest(const char *name, enum gcry_md_algos *algo) {
 	int i;
 
 	for(i = 0; i < sizeof(digesttable) / sizeof(*digesttable); i++) {
@@ -47,10 +47,8 @@ static bool nametodigest(const char *name, int *algo) {
 	return false;
 }
 
-static bool nidtodigest(int nid, int *algo) {
-	int i;
-
-	for(i = 0; i < sizeof(digesttable) / sizeof(*digesttable); i++) {
+static bool nidtodigest(int nid, enum gcry_md_algos *algo) {
+	for(int i = 0; i < sizeof(digesttable) / sizeof(*digesttable); i++) {
 		if(nid == digesttable[i].nid) {
 			*algo = digesttable[i].algo;
 			return true;
@@ -60,10 +58,8 @@ static bool nidtodigest(int nid, int *algo) {
 	return false;
 }
 
-static bool digesttonid(int algo, int *nid) {
-	int i;
-
-	for(i = 0; i < sizeof(digesttable) / sizeof(*digesttable); i++) {
+static bool digesttonid(enum gcry_md_algos algo, int *nid) {
+	for(int i = 0; i < sizeof(digesttable) / sizeof(*digesttable); i++) {
 		if(algo == digesttable[i].algo) {
 			*nid = digesttable[i].nid;
 			return true;
@@ -73,7 +69,7 @@ static bool digesttonid(int algo, int *nid) {
 	return false;
 }
 
-static bool digest_open(digest_t *digest, int algo, int maclength) {
+static bool digest_open(digest_t *digest, enum gcry_md_algos algo, size_t maclength) {
 	if(!digesttonid(algo, &digest->nid)) {
 		logger(DEBUG_ALWAYS, LOG_DEBUG, "Digest %d has no corresponding nid!", algo);
 		return false;
@@ -93,8 +89,8 @@ static bool digest_open(digest_t *digest, int algo, int maclength) {
 	return true;
 }
 
-bool digest_open_by_name(digest_t *digest, const char *name, int maclength) {
-	int algo;
+bool digest_open_by_name(digest_t *digest, const char *name, size_t maclength) {
+	enum gcry_md_algos algo;
 
 	if(!nametodigest(name, &algo)) {
 		logger(DEBUG_ALWAYS, LOG_DEBUG, "Unknown digest name '%s'!", name);
@@ -104,8 +100,8 @@ bool digest_open_by_name(digest_t *digest, const char *name, int maclength) {
 	return digest_open(digest, algo, maclength);
 }
 
-bool digest_open_by_nid(digest_t *digest, int nid, int maclength) {
-	int algo;
+bool digest_open_by_nid(digest_t *digest, int nid, size_t maclength) {
+	enum gcry_md_algos algo;
 
 	if(!nidtodigest(nid, &algo)) {
 		logger(DEBUG_ALWAYS, LOG_DEBUG, "Unknown digest ID %d!", nid);
@@ -115,7 +111,7 @@ bool digest_open_by_nid(digest_t *digest, int nid, int maclength) {
 	return digest_open(digest, algo, maclength);
 }
 
-bool digest_open_sha1(digest_t *digest, int maclength) {
+bool digest_open_sha1(digest_t *digest, size_t maclength) {
 	return digest_open(digest, GCRY_MD_SHA1, maclength);
 }
 
@@ -163,7 +159,7 @@ bool digest_create(digest_t *digest, const void *indata, size_t inlen, void *out
 }
 
 bool digest_verify(digest_t *digest, const void *indata, size_t inlen, const void *cmpdata) {
-	ssize_t len = digest->maclength;
+	size_t len = digest->maclength;
 	uint8_t outdata[len];
 
 	return digest_create(digest, indata, inlen, outdata) && !memcmp(cmpdata, outdata, len);
diff --git a/src/gcrypt/digest.h b/src/gcrypt/digest.h
index 6ba5dd56..9a9c611e 100644
--- a/src/gcrypt/digest.h
+++ b/src/gcrypt/digest.h
@@ -25,15 +25,15 @@
 #define DIGEST_MAX_SIZE 64
 
 typedef struct digest {
-	int algo;
+	enum gcry_md_algos algo;
 	int nid;
-	ssize_t maclength;
+	size_t maclength;
 	gcry_md_hd_t hmac;
 } digest_t;
 
-extern bool digest_open_by_name(struct digest *, const char *name, int maclength);
-extern bool digest_open_by_nid(struct digest *, int nid, int maclength);
-extern bool digest_open_sha1(struct digest *, int maclength);
+extern bool digest_open_by_name(struct digest *, const char *name, size_t maclength);
+extern bool digest_open_by_nid(struct digest *, int nid, size_t maclength);
+extern bool digest_open_sha1(struct digest *, size_t maclength);
 extern void digest_close(struct digest *);
 extern bool digest_create(struct digest *, const void *indata, size_t inlen, void *outdata);
 extern bool digest_verify(struct digest *, const void *indata, size_t inlen, const void *digestdata);
diff --git a/src/net_setup.c b/src/net_setup.c
index 90a78495..02dfd74b 100644
--- a/src/net_setup.c
+++ b/src/net_setup.c
@@ -863,6 +863,8 @@ static bool setup_myself(void) {
 	int compression = COMPRESS_NONE;
 
 	if(get_config_int(lookup_config(config_tree, "Compression"), &compression)) {
+		myself->incompression = compression;
+
 		switch(myself->incompression) {
 		case COMPRESS_LZ4:
 #ifdef HAVE_LZ4
@@ -908,8 +910,6 @@ static bool setup_myself(void) {
 			logger(DEBUG_ALWAYS, LOG_ERR, "Compression level %i is unrecognized by this node.", myself->incompression);
 			return false;
 		}
-
-		myself->incompression = compression;
 	} else {
 		myself->incompression = COMPRESS_NONE;
 	}
diff --git a/src/openssl/digest.c b/src/openssl/digest.c
index 2be310d2..82364e71 100644
--- a/src/openssl/digest.c
+++ b/src/openssl/digest.c
@@ -27,13 +27,13 @@
 #include "../digest.h"
 #include "../logger.h"
 
-static digest_t *digest_open(const EVP_MD *evp_md, ssize_t maclength) {
+static digest_t *digest_open(const EVP_MD *evp_md, size_t maclength) {
 	digest_t *digest = xzalloc(sizeof(*digest));
 	digest->digest = evp_md;
 
 	size_t digestlen = EVP_MD_size(digest->digest);
 
-	if(maclength < 0 || (size_t) maclength > digestlen) {
+	if(maclength == DIGEST_ALGO_SIZE || maclength > digestlen) {
 		digest->maclength = digestlen;
 	} else {
 		digest->maclength = maclength;
@@ -42,7 +42,7 @@ static digest_t *digest_open(const EVP_MD *evp_md, ssize_t maclength) {
 	return digest;
 }
 
-digest_t *digest_open_by_name(const char *name, int maclength) {
+digest_t *digest_open_by_name(const char *name, size_t maclength) {
 	const EVP_MD *evp_md = EVP_get_digestbyname(name);
 
 	if(!evp_md) {
@@ -53,7 +53,7 @@ digest_t *digest_open_by_name(const char *name, int maclength) {
 	return digest_open(evp_md, maclength);
 }
 
-digest_t *digest_open_by_nid(int nid, ssize_t maclength) {
+digest_t *digest_open_by_nid(int nid, size_t maclength) {
 	const EVP_MD *evp_md = EVP_get_digestbynid(nid);
 
 	if(!evp_md) {
diff --git a/src/openssl/prf.c b/src/openssl/prf.c
index b7a77f99..62770859 100644
--- a/src/openssl/prf.c
+++ b/src/openssl/prf.c
@@ -30,7 +30,7 @@
  */
 
 static bool prf_xor(int nid, const uint8_t *secret, size_t secretlen, uint8_t *seed, size_t seedlen, uint8_t *out, size_t outlen) {
-	digest_t *digest = digest_open_by_nid(nid, -1);
+	digest_t *digest = digest_open_by_nid(nid, DIGEST_ALGO_SIZE);
 
 	if(!digest) {
 		return false;
diff --git a/src/protocol_auth.c b/src/protocol_auth.c
index c5ba071d..e16fe4fc 100644
--- a/src/protocol_auth.c
+++ b/src/protocol_auth.c
@@ -518,13 +518,9 @@ bool send_metakey(connection_t *c) {
 		c->outcipher = cipher_open_by_name("aes-256-cfb");
 	}
 
-	if(!c) {
-		return false;
-	}
-
 	c->outbudget = cipher_budget(c->outcipher);
 
-	if(!(c->outdigest = digest_open_by_name("sha256", -1))) {
+	if(!(c->outdigest = digest_open_by_name("sha256", DIGEST_ALGO_SIZE))) {
 		return false;
 	}
 
@@ -639,7 +635,7 @@ bool metakey_h(connection_t *c, const char *request) {
 	c->inbudget = cipher_budget(c->incipher);
 
 	if(digest) {
-		if(!(c->indigest = digest_open_by_nid(digest, -1))) {
+		if(!(c->indigest = digest_open_by_nid(digest, DIGEST_ALGO_SIZE))) {
 			logger(DEBUG_ALWAYS, LOG_ERR, "Error during initialisation of digest from %s (%s)", c->name, c->hostname);
 			return false;
 		}
diff --git a/src/protocol_key.c b/src/protocol_key.c
index 65fd35ec..012f2267 100644
--- a/src/protocol_key.c
+++ b/src/protocol_key.c
@@ -356,7 +356,7 @@ bool send_ans_key(node_t *to) {
 
 	if(myself->indigest) {
 		to->indigest = digest_open_by_nid(digest_get_nid(myself->indigest),
-		                                  (ssize_t)digest_length(myself->indigest));
+		                                  digest_length(myself->indigest));
 
 		if(!to->indigest) {
 			abort();
@@ -397,11 +397,11 @@ bool ans_key_h(connection_t *c, const char *request) {
 	char address[MAX_STRING_SIZE] = "";
 	char port[MAX_STRING_SIZE] = "";
 	int cipher, digest;
-	ssize_t maclength;
+	size_t maclength;
 	compression_level_t compression;
 	node_t *from, *to;
 
-	if(sscanf(request, "%*d "MAX_STRING" "MAX_STRING" "MAX_STRING" %d %d %zd %d "MAX_STRING" "MAX_STRING,
+	if(sscanf(request, "%*d "MAX_STRING" "MAX_STRING" "MAX_STRING" %d %d %zu %d "MAX_STRING" "MAX_STRING,
 	                from_name, to_name, key, &cipher, &digest, &maclength,
 	                &compression, address, port) < 7) {
 		logger(DEBUG_ALWAYS, LOG_ERR, "Got bad %s from %s (%s)", "ANS_KEY", c->name,
@@ -467,7 +467,7 @@ bool ans_key_h(connection_t *c, const char *request) {
 	}
 
 	switch(compression) {
-	case 12:
+	case COMPRESS_LZ4:
 #ifdef HAVE_LZ4
 		break;
 #else
@@ -476,8 +476,8 @@ bool ans_key_h(connection_t *c, const char *request) {
 		return true;
 #endif
 
-	case 11:
-	case 10:
+	case COMPRESS_LZO_HI:
+	case COMPRESS_LZO_LO:
 #ifdef HAVE_LZO
 		break;
 #else
@@ -486,15 +486,15 @@ bool ans_key_h(connection_t *c, const char *request) {
 		return true;
 #endif
 
-	case 9:
-	case 8:
-	case 7:
-	case 6:
-	case 5:
-	case 4:
-	case 3:
-	case 2:
-	case 1:
+	case COMPRESS_ZLIB_9:
+	case COMPRESS_ZLIB_8:
+	case COMPRESS_ZLIB_7:
+	case COMPRESS_ZLIB_6:
+	case COMPRESS_ZLIB_5:
+	case COMPRESS_ZLIB_4:
+	case COMPRESS_ZLIB_3:
+	case COMPRESS_ZLIB_2:
+	case COMPRESS_ZLIB_1:
 #ifdef HAVE_ZLIB
 		break;
 #else
@@ -503,7 +503,7 @@ bool ans_key_h(connection_t *c, const char *request) {
 		return true;
 #endif
 
-	case 0:
+	case COMPRESS_NONE:
 		break;
 
 	default:
```

src/compression.h
```diff
@@ -0,0 +1,25 @@
...
+	COMPRESS_ZLIB_8 = 8,
+	COMPRESS_ZLIB_9 = 9,
+
+	COMPRESS_LZO_LO = 10,
+	COMPRESS_LZO_HI = 11,
+
+	COMPRESS_LZ4 = 12,
+
+	COMPRESS_GUARD = INT_MAX, /* ensure that sizeof(compression_level_t) == sizeof(int) */
+} compression_level_t;
```
gsliepen, line 23: I'm getting compiler warnings myself now:

```
control.c:133:29: warning: format ‘%d’ expects argument of type ‘int *’, but argument 3 has type ‘compression_level_t *’ [-Wformat=]
  133 |   sscanf(request, "%*d %*d %d", &c->outcompression);
      |                            ~^   ~~~~~~~~~~~~~~~~~~
      |                             |   |
      |                             |   compression_level_t *
      |                             int *
```

And:

```
protocol_key.c:404:21: warning: format ‘%d’ expects argument of type ‘int *’, but argument 9 has type ‘compression_level_t *’ [-Wformat=]
  404 |  if(sscanf(request, "%*d "MAX_STRING" "MAX_STRING" "MAX_STRING" %d %d %zu %d "MAX_STRING" "MAX_STRING,
      |                     ^~~~~~
```

So for `printf()` and `scanf()`, we can't dereference pointers to `enum`s. I like the `enum` and I think we should keep it, but maybe revert the type of the variables storing compression levels back to `int`.

gsliepen (changes requested):

src/compression.h
hg: Rats. For some reason, neither clang nor gcc were reporting these warnings on my system.

I reverted compression fields to `int` and added `-Werror` to the [static](https://github.com/gsliepen/tinc/runs/3218917811?check_suite_focus=true#step:12:855) analysis [job](https://github.com/hg/tinc/runs/3218249344?check_suite_focus=true) so as not to burden you with such trivia in the future. We probably won't be able to add it everywhere because of [old compilers](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53119) that nag about `struct type_t xxx = { 0 };`, and macOS which deprecates something every release.

Failing on `-Werror` won't affect other checks.

src/route.c
```diff
@@ -587,7 +587,7 @@ static void fragment_ipv4_packet(node_t *dest, vpn_packet_t *packet, length_t et
 	todo = ntohs(ip.ip_len) - ip_size;
 
 	if(ether_size + ip_size + todo != packet->len) {
-		logger(DEBUG_TRAFFIC, LOG_WARNING, "Length of packet (%d) doesn't match length in IPv4 header (%d)", packet->len, (int)(ether_size + ip_size + todo));
+		logger(DEBUG_TRAFFIC, LOG_WARNING, "Length of packet (%d) doesn't match length in IPv4 header (%lu)", packet->len, ether_size + ip_size + todo);
```
gsliepen, line 590: Should be `%zu` instead of `%lu`, since it's a `size_t`?

gsliepen (changes requested):
I think that one %lu is the only thing left :)

src/route.c
hg: Makes sense, thanks. To be honest, I am more used to linters for more high level languages, which do not produce so many broken auto"fixes". Here you cannot really rely on what the linter recommends. I'll double check it in the future.

The static analysis job here failed to checkout code (GitHub Actions are not particularly reliable), so here's a second [test run](https://github.com/hg/tinc/actions/runs/1089778614).

Issue #289, opened 2021-07-20
Weekly / Nightly build


fangfufu:
Since we are doing CI. Why don't we generate a weekly or even nightly build? 
(Tagging you, @hg, since you have been working on CI :slightly_smiling_face: )

hg:
Sure, I also find this useful. Most of the required infrastructure is already there. RPM build configs will have to be ported first (probably ripped out from Fedora packages), and whatever weirdness Windows requires to create installers.

No idea about Macs with their notarization requirements. They already have [this](https://github.com/Homebrew/homebrew-core/blob/2684724e2adda0c0a220481e58b1bb69e42d7a6c/Formula/tinc.rb) and it seems to be the preferred way to install third-party applications, so maybe Mac can be skipped altogether.

I would like to finish the meson port first, though, so as not to repeat the same work twice. (If it's accepted.)

fangfufu:
Done at https://github.com/gsliepen/tinc/commit/7619f84393f824a84e20e03167304b27d3581ac0

PR #299, opened 2021-08-08
CI: cross-compilation for MIPS/armv7; build packages on every push

hg:
I propose a few more changes to CI.

Tests:

- https://github.com/hg/tinc/actions/runs/1110145524
- https://github.com/hg/tinc/actions/runs/1110181726
- https://github.com/hg/tinc/actions/runs/1110058391
- https://github.com/hg/tinc/actions/runs/1110104752

# build more packages

re #289: build deb + rpm + windows packages on every push to the main branch, and on every release:

- deb:
  - Debian stable
  - Debian testing
  - Ubuntu LTS (latest)
- rpm:
  - RHEL-ish 7
  - RHEL-ish 8
  - Fedora (latest)
- and a simple installer for Windows

@fangfufu: did you have any other distributions in mind?

The packages are published as a pre-release on [this page](https://github.com/hg/tinc/releases/tag/latest), or as a [proper release](https://github.com/hg/tinc/releases/tag/release-1.75) if you use a tag.

The pre-release will always reuse the same tag `latest`. GitHub does not allow creating releases without tags, so there's not much choice here.

# cross-compilation

Add a couple of cross-compilation jobs for two architectures frequently seen in cheap routers: mips and armv7; run tests on resulting binaries using qemu user virtualization.

mips also gives us additional testing on a big-endian architecture.

I also wanted to add s390 (because of [this](https://github.com/alpinelinux/aports/blob/5e5504e2556b0b6fa71c3e29ad997c87d46d7697/community/tinc-pre/APKBUILD#L9-L11)), but it kept failing on key generation with `SIGILL` (seems to be an omission in qemu).

# cleanups

Reuse the same configuration flags and package list as much as possible (by moving them out of the yaml config).

Since the test suite seems to be stable now, I reduced the number of Windows and macOS jobs to one, which is now using the same test script as the Linux jobs (and runs tests both with the default config, and without the legacy protocol). There's not much sense in running multiple jobs: macOS is quickly deprecated after the next release is available, and Windows 2016/2019 are very similar (both use the same msys2 environment).


fangfufu:
Should we add the latest Ubuntu (21.04) as well?

ci/deps.sh
```diff
@@ -0,0 +1,86 @@
...
+set -eu
+
+deps_alpine() {
+  apk upgrade
+
+  apk add \
+    git binutils make autoconf automake gcc linux-headers diffutils texinfo \
+    procps socat shadow sudo openssl-dev zlib-dev lzo-dev ncurses-dev \
+    readline-dev musl-dev lz4-dev vde2-dev
+}
```
fangfufu, line 12: What are these Alpine stuff for? Just wondering.

hg: It's the same as for the other distributions — to build the project and run the test suite. We don't build packages for it though, because Alpine is updated very frequently anyway.

hg:
> Should we add the latest Ubuntu (21.04) as well?

Makes sense, added. The tag misleadingly says "rolling", but in reality it's the latest official release.

---

21.04 doesn't seem to ship `libvdeplug_dyn.h`, just like Debian sid. Tinc will probably have to be updated to support this, I'll look into it later.

ci/deps.sh
fangfufu: In that case, could we provide an Alpine package please? I had a look at their repository, they don't seem to provide a package for tinc 1.1. 

This is the link I looked at: 
https://pkgs.alpinelinux.org/packages?name=tinc&branch=edge&repo=main

hg: It's there, but is named `tinc-pre` and is located in the `community` repository.

https://pkgs.alpinelinux.org/packages?name=tinc-pre&branch=edge&repo=community

gsliepen:
> 21.04 doesn't seem to ship libvdeplug_dyn.h, just like Debian sid. Tinc will probably have to be updated to support this, I'll look into it later.

I created issue #300 for this.

hg:
This will add a small amount of work to package maintainers, but I think it's time to rename README to README.md to force GitHub and Gitlab to properly render it as a Markdown document.

The releases section is simply not discoverable. Nobody will know about nightly builds unless we direct users to it with big, bold letters:

https://github.com/hg/tinc/tree/ci-test#nightly-builds

Should I push this here?

gsliepen:
> Should I push this here?

Sure!

gsliepen (commented):
The scripts in the `ci/` directory should never be runi by users directly. It might help to add a README.md in that directory explaining their purpose. Maybe also rename the directory to `.ci` so it's hidden by default?

ci/deps.sh
fangfufu: Ah ok. :)

hg:
>The scripts in the ci/ directory should never be runi by users directly. It might help to add a README.md in that directory explaining their purpose. Maybe also rename the directory to .ci so it's hidden by default?

I did both. We could keep them in `.github/`, but the paths are unmanageable, and (strictly speaking) it's a directory for GitHub magic files, so who knows what actions they could auto-trigger in the future.

vde2 is now used on every distribution except RHEL-based ones (where you have to build it from source).

Issue #300, opened 2021-08-08
Add support for newer versions of libvdeplug.


gsliepen:
Tinc fails to compile on Debian unstable and Ubuntu 21.04 with `--enable-vde`, the following error message occurs during the `./configure` step:

```
checking libvdeplug_dyn.h usability... no
checking libvdeplug_dyn.h presence... no
checking for libvdeplug_dyn.h... no
configure: error: VDE plug header files not found.
```

PR #302, opened 2021-08-08
Use libvdeplug.h instead of libvdeplug_dyn.h

fangfufu:
This is my attempt to address issue https://github.com/gsliepen/tinc/issues/300. I changed include from ``libvdeplug_dyn.h`` rather than ``libvdeplug.h``. 

The ``libvdeplug.h`` from Debian Unstable is almost identical to the one from Debian Buster. ~~I think using the statically linked version of ``libvdeplug`` might be the quickest and easiest fix.~~ (This is still dynamic linking, please refer to https://github.com/gsliepen/tinc/pull/302#issuecomment-894950250)

### DIFF
--- a/src/vde_device.c
+++ b/src/vde_device.c
@@ -19,7 +19,7 @@
 
 #include "system.h"
 
-#include <libvdeplug_dyn.h>
+#include <libvdeplug.h>
 
 #include "conf.h"
 #include "device.h"
@@ -29,20 +29,12 @@
 #include "route.h"
 #include "xalloc.h"
 
-static struct vdepluglib plug;
 static struct vdeconn *conn = NULL;
 static int port = 0;
 static char *group = NULL;
 static const char *device_info = "VDE socket";
 
 static bool setup_device(void) {
-	libvdeplug_dynopen(plug);
-
-	if(!plug.dl_handle) {
-		logger(DEBUG_ALWAYS, LOG_ERR, "Could not open libvdeplug library!");
-		return false;
-	}
-
 	if(!get_config_string(lookup_config(config_tree, "Device"), &device)) {
 		xasprintf(&device, RUNSTATEDIR "/vde.ctl");
 	}
@@ -59,14 +51,14 @@ static bool setup_device(void) {
 		.mode = 0700,
 	};
 
-	conn = plug.vde_open(device, identname, &args);
+	conn = vde_open(device, identname, &args);
 
 	if(!conn) {
 		logger(DEBUG_ALWAYS, LOG_ERR, "Could not open VDE socket %s", device);
 		return false;
 	}
 
-	device_fd = plug.vde_datafd(conn);
+	device_fd = vde_datafd(conn);
 
 #ifdef FD_CLOEXEC
 	fcntl(device_fd, F_SETFD, FD_CLOEXEC);
@@ -83,12 +75,8 @@ static bool setup_device(void) {
 
 static void close_device(void) {
 	if(conn) {
-		plug.vde_close(conn);
+		vde_close(conn);
 		conn = NULL;
-	}
-
-	if(plug.dl_handle) {
-		libvdeplug_dynclose(plug);
 	}
 
 	free(device);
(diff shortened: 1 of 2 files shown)


fangfufu:
@hg, what do you think of this?

Nable80:
> non-dynamic loading

Is it just static linking?

> ```c
> 	ssize_t lenin = (ssize_t) vde_recv(conn, DATA(packet), MTU, 0);
> ...
> 	if((ssize_t)vde_send(conn, DATA(packet), packet->len, 0) < 0) {
> ```

https://manpages.debian.org/unstable/libvdeplug-dev/vde_recv.3.en.html tells me that ``vde_recv`` and ``vde_send`` already return ``ssize_t``, why are these casts needed?

fangfufu:
I think it is basically static linking. My knowledge of the linker isn't great. I have never personally written a dynamically linked C program. I basically just worded it in a conservative way. You are write that ``(ssize_t)`` casting is no longer required. It was required due to dynamic linking. I have amended the commit.

Nable80:
You can use ``ldd`` tool to see which libraries are loaded dynamically when the binary is executed, e.g.:
```
ldd ./tincd
        linux-vdso.so.1 (0x00007fff837e1000)
        libcrypto.so.1.1 => /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1 (0x00007fa53e40d000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fa53e3f0000)
        liblzo2.so.2 => /lib/x86_64-linux-gnu/liblzo2.so.2 (0x00007fa53e3cb000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fa53e206000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fa53e200000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fa53e1de000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fa53e7bf000)
```

FYI: statically-linked libraries won't appear in this list. Maybe libvde is still linked dynamically after your changes (and it's not a problem) but it happens when process is started (regular dynamic library behavior) instead of later run-time (plugin style).

hg:
It still uses dynamic linking. It's just that the task of linking the library is passed to the system dynamic linker at tincd start time, instead of doing it manually with `libvdeplug_dynopen` when vde functionality is actually needed. This is probably the best solution in any case IMHO.

I tested it on [Ubuntu 21.04 and Debian Sid](https://github.com/hg/tinc/actions/runs/1111641737), and manually on FreeBSD 11, NetBSD 8, and DragonflyBSD 6. Works fine.

This should probably go in before #299, I will need to adjust build scripts some more.

---

> ... tells me that vde_recv and vde_send already return ssize_t, why are these casts needed?

It made sense for the old header because of a bug in declarations there. See [this](https://github.com/gsliepen/tinc/pull/296#discussion_r680519175). This header seems to be fine everywhere I checked.

src/vde_device.c
```diff
@@ -101,7 +89,7 @@ static void close_device(void) {
 }
 
 static bool read_packet(vpn_packet_t *packet) {
-	ssize_t lenin = (ssize_t) plug.vde_recv(conn, DATA(packet), MTU, 0);
+	ssize_t lenin = vde_recv(conn, DATA(packet), MTU, 0);
```
gsliepen, line 92: I just see now that according to the manpage, anything less than 14 bytes should be discarded.

fangfufu:
By the way, @gsliepen / @hg, which linter command should I be using?

hg:
I use compilers with sanitizers (both clang and gcc), clang-tidy, and cppcheck (which breaks on some things because it uses its own parser instead of a compiler backend like clang-tidy does, but it's still useful).

Use this as a starting point:

https://github.com/gsliepen/tinc/blob/bf7abd5a059d27dc544b435b67f7f1d5419d318b/.github/workflows/test.yml#L63-L68

and

https://github.com/gsliepen/tinc/tree/1.1/.github/workflows/sanitizers

hg:
Ah, yes, and you need this to reformat C

    $ astyle -r --options=.astylerc  --formatted '*.c' '*.h'

and 

    $ shellcheck -x script.sh
    $ shfmt -d -i 2 -s script.sh

if  any scripts were changed.

src/vde_device.c
fangfufu: I added some more message logging below. Please do let me know if ``event_exit()`` should be called in those cases. I feel they are recoverable errors, so I decided not to call it.

fangfufu:
Ah apologies. I meant formatter rather than linter. I will leave linting / santizing to you CI scripts. Thanks for recognising that I was probably confused. :slightly_smiling_face:

gsliepen:
You can run `make astyle` from the top level build directory. Maybe I should add a "precommit" rule or something like that that calls a bunch of stuff to reformat and do some simple checks.

fangfufu:
I am pretty sure you have to set up pre-commit hooks manually after you clone the repository. But I suppose it would be nice to provide a script for everyone to use, that can be inside the repository. You can symlink the scrpt as a pre-commit hook.

src/vde_device.c
gsliepen: I think those are recoverable as well. Anyway, there is already some code that will cause the event loop to exit if `read_packet()` returns `false` many times in succession.

PR #304, opened 2021-08-10
Reduce pointer indirection for global list_t variables

fangfufu:
Converted cmdline_conf, connection_list, outgoing_list from
pointer-to-structs to structs.

Created list_empty_list for these structs. This is necessary,
because list_delete_list frees the supplied list_t pointer.

Part of https://github.com/gsliepen/tinc/issues/294

Issue #294, opened 2021-07-29
Reduce pointer indirection


gsliepen:
There are several global and member variables which are pointers to structs that could be replaced by those structs themselves. This avoids having to allocate and free memory for them, and might improve performance a little bit. Candidates are:

- `splay_tree_t`s
- `list_t`s
- `cipher_t` and `digest_t`s in `connection_t`, as they are likely to be used

Futhermore, in `node_t`, we could group all the variables that are only needed when we actually want to exchange data with that node (packets counters, ciphers, digests, probe state) into a separate struct, and have a single pointer to that new struct.

fangfufu:
Okay, apologies, in future I will make sure I finalise the commit message before adding the link to Github issue. I am too used to Gerrit. Sorry for polluting the conversation thread here.

fangfufu:
Changing this to enhancement, as it improves performance.

PR #307, opened 2021-08-10
Replace pointers to global splay trees with structs.

hg:
Related to #294

This gets rid of global pointers to splay trees.

In addition to this, I also wanted to replace pointers in struct members, but it's more difficult and will take some time. In light of #304 I'll PR this now so we could avoid repeating the same work.


src/splay_tree.c
```diff
@@ -604,12 +604,22 @@ void splay_delete(splay_tree_t *tree, void *data) {
 
 /* Fast tree cleanup */
 
-void splay_delete_tree(splay_tree_t *tree) {
+void splay_free_nodes(splay_tree_t *tree) {
 	for(splay_node_t *node = tree->head, *next; node; node = next) {
 		next = node->next;
 		splay_free_node(tree, node);
 	}
+}
```
gsliepen, line 612: This is a potential landmine; it doesn't zero `tree->head` and related pointers. I also don't see why we need both `splay_free_nodes()` and the slower `splay_delete_nodes()`. If we are going to remove all the nodes we don't need to unlink them first. I think it's better to create one function named `splay_empty_tree()`, like in @fangfufu's patch. Call `splay_free_nodes()` and then set `tree->head`, `tree->tail` and `tree->root` all to `NULL`.

gsliepen (changes requested):

src/splay_tree.c
gsliepen: Also `tree->count = 0`, and `tree->generation++`.

hg: That's much better. Thank you.

I am not sure if it makes sense to inline the remaining two configuration trees in `connection_t` and `address_cache_t`. They are lazily initialized and (I think) can be reused many times over. Initialization relies on some data that is not known when the structure itself is allocated (like node name). This can be circumvented with additional flags, etc, but it seems to only make code worse than it is now.

https://github.com/gsliepen/tinc/blob/0871c3095151bce6a4031a2662aa51b7193b855c/src/connection.h#L107-L108

https://github.com/gsliepen/tinc/blob/95230693d2fda7bde6d44aea170ae12caf8eab7a/src/address_cache.h#L28-L30

src/splay_tree.c
```diff
@@ -604,12 +604,25 @@ void splay_delete(splay_tree_t *tree, void *data) {
...
+	splay_free_nodes(tree);
+
+	tree->head = NULL;
+	tree->tail = NULL;
+	tree->root = NULL;
+	tree->count = 0;
+	tree->generation++;
+}
+
+void splay_free_nodes(splay_tree_t *tree) {
```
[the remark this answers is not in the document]
hg: I threw it away altogether, less cognitive load in exchange for tiny performance hit when freeing configuration.

PR #309, opened 2021-08-10
Test pull request to see if windows CI works (do not merge) 

fangfufu:
**Do not merge**

I can't seem to trigger CI runs on my own forks, without creating a pull request.

### DIFF
--- a/src/autoconnect.c
+++ b/src/autoconnect.c
@@ -53,7 +53,7 @@ static void make_new_connection() {
 
 		bool found = false;
 
-		for list_each(outgoing_t, outgoing, outgoing_list) {
+		for list_each(outgoing_t, outgoing, &outgoing_list) {
 			if(outgoing->node == n) {
 				found = true;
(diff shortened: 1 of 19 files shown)


fangfufu:
It appears that CI is broken right now... The Github workflow isn't running at all.

gsliepen:
It's because you pushed exactly the same commit as in #308. I think GitHub notices it's not a new commit hash and thus skips checks.

fangfufu:
@gsliepen , I created this new pull request because the old one was only running the workflows outside of Github. I have just created a new commit with a dummy file, and it is still not running the workflows from Github itself.

fangfufu:
I recommend continuing the discussion of Github workflows not running at https://github.com/gsliepen/tinc/issues/310

fangfufu:
Okay, Github workflow issue was due to Github infrastructure. I am still investigating the problem with Windows CI workflow.

PR #316, opened 2021-08-14
hash table fix

splitice:
1. Clears keys during hash_clear
2. Explicitly clears memory during subnet init (bug introduced during merge)
3. Fixes memcmp read past bounds (read size 8, not 4/6/8)


splitice:
Note: Not yet real world tested. Currently setting up a new test environment (with OpenSSL 1.1.1)

Issue #306, opened 2021-08-10
Windows CI checks fail due to timeout. 


fangfufu:
The Windows CI checks (``windows`` and ``windows --disable-legacy-protocol``) have been failing due to exceeding the maximum execution time of 20 minutes. 

I had a look around, I couldn't find an issue or a pull request dedicated to this. So I am opening a new issue. 

Tagging @hg

hg:
TBH I have no idea at the moment what the issue might be. See:

https://github.com/gsliepen/tinc/commit/e050753d1fc4d4465032919746c2a4db78f19932#comments

fangfufu:
I am just wondering if we should turn the Windows CI workflow off for now. They aren't doing much right now. If we turn them off, we can get green ticks. 

What do you think, @gsliepen?

fangfufu:
> TBH I have no idea at the moment what the issue might be. See:
> 
> [e050753#comments](https://github.com/gsliepen/tinc/commit/e050753d1fc4d4465032919746c2a4db78f19932#comments)

I can confirm this causes the problem, have a look at the run result of https://github.com/fangfufu/tinc/commit/5b3ed1540f385e708eee9ff025bfdd52f80665a1. I am not sure why it causes the problem either. I will try and investigate further.

fangfufu:
I managed to replicate the same problem by running the CI script locally on my Windows 10 machine. I have no idea what might have caused it either. This is a bit beyond me.

fangfufu:
@gsliepen, @hg, if you run the whole CI script under mingw32 toolchain, it works fine. So this problem is compiler specific.

hg:
Yeah, this particular issue is triggered by `scripts.test`. `bar` repeatedly sends an Ethernet frame to `foo` with the type field set to zero. See this with wireshark:

[dump.zip](https://github.com/gsliepen/tinc/files/6973857/dump.zip)

It's easy to un-hang by assigning this field as is done elsewhere in the project:

https://github.com/hg/tinc/commit/fa7e7f42e642ad54f1c8b633506fbc5769c71139

the test suite passes just fine

- https://github.com/hg/tinc/runs/3301782019
- https://github.com/hg/tinc/runs/3301518716

but then hangs anyway after running the last test.

I'll dig into it later, I need to finish some other things first.

fangfufu:
I managed to run this under MSYS2 MINGW32. This is my test folder. Do let me know if you want more things.

[...]

Issue #315, opened 2021-08-14
Sanitizer (address) CI checks fail due to timeout. 


fangfufu:
The sanitizer (address) CI checks have been failing due to exceeding the maximum execution time. Raising the execution time limit has not helped. 

I believe the offending commit is https://github.com/gsliepen/tinc/commit/9a018c2e371eb1cef9708ac71653f2f2868895fa. There have been two attempts to raise the execution time limit - https://github.com/gsliepen/tinc/commit/0891154941b4f638b8a974ad5085e130a451ceec to 20 minutes, and https://github.com/fangfufu/tinc/commit/202086ae1374a9ad620422b1f137dee1d7869e1b to 40 minutes. They both have not resolved the issue. 

Tagging @splitice , @gsliepen

fangfufu:
I guess the question is whether this actually affects real life use...

hg:
Sure.

```
tincd 1.1pre18-87-g08911549 (Aug 14 2021 20:49:21) starting, debug level 0
Error reading RSA private key file `/etc/tinc/work//rsa_key.priv': No such file or directory
Create an RSA key pair with `tinc -n . generate-rsa-keys'.
Support for legacy protocol disabled.
/dev/net/tun is a Linux tun/tap device (tun mode)
Ready
=================================================================
==832647==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffe07413964 at pc 0x5585a0d09652 bp 0x7ffe07413770 sp 0x7ffe07412f20
READ of size 7 at 0x7ffe07413964 thread T0
    #0 0x5585a0d09651 in MemcmpInterceptorCommon(void*, int (*)(void const*, void const*, unsigned long), void const*, void const*, unsigned long) (src/tincd+0x15c651)
    #1 0x5585a0d09cc8 in memcmp (src/tincd+0x15ccc8)
    #2 0x5585a0e73ee4 in hash_search_ipv4_t src/subnet.c:105:1
    #3 0x5585a0e77486 in lookup_subnet_ipv4 src/subnet.c:256:10
    #4 0x5585a0e369ed in route_ipv4 src/route.c:632:11
    #5 0x5585a0e34460 in route src/route.c:1156:4
    #6 0x5585a0dc3b33 in handle_device_data src/net_packet.c:1910:3
    #7 0x5585a0d80bd9 in event_loop src/event.c:355:5
    #8 0x5585a0da552d in main_loop src/net.c:505:6
[...]

PR #317, opened 2021-08-14
Improve failure detection in the test suite.

hg:
Improve detection of tinc failures in the test suite — it should now report failures as failures instead of hanging indefinitely.

Prompted by #315.

The reason behind `--foreground` was to put the time limit only on the, em, foreground process, ignoring anything that it might start and leave running in the background (like tinc). After rewriting the test library so many times I don't believe it's actually needed anymore, any it's best to throw it away in any case because it's not supported by most `timeout`s.

Things that should have failed (and did):

- https://github.com/hg/tinc/runs/3329650122
- https://github.com/hg/tinc/runs/3329650316

---

A bit of a rant: it's way too late for that, but after bashing my head against so many pointless differences in Unix utilities I am of the opinion that the test suite for cross-platform projects should be written in something like Python, or maybe even C. Diversity is good when there is actual diversity, but not when you have ten sets of utilities that are pretty much the same, except for tiny little differences like: this `tail` does X on SIGPIPE, that `tail` doesn't; this `wc` prints whitespace around the number, that one doesn't, this shell waits for all child jobs to finish, that one doesn't, and so on and so forth.

Autotools developers deserve a monument for their efforts.

### DIFF
--- a/test/compression.test
+++ b/test/compression.test
@@ -106,13 +106,13 @@ for level in $levels; do
   wait_script foo hosts/bar-up
   wait_script bar hosts/foo-up
 
-  ip netns exec foo \
-    socat -u TCP4-LISTEN:$recv_port_foo,reuseaddr OPEN:"$tmp_file",creat &
+  try_limit_time 60 sh <<EOF
+    set -eu
+    ip netns exec foo socat -u TCP4-LISTEN:$recv_port_foo,reuseaddr OPEN:"$tmp_file",creat &
+    ip netns exec bar socat -u OPEN:"$ref_file" TCP4:$ip_foo:$recv_port_foo,retry=30 &
+    wait
+EOF
 
-  ip netns exec bar \
-    socat -u OPEN:"$ref_file" TCP4:$ip_foo:$recv_port_foo,retry=30 &
-
-  wait
   diff -w "$ref_file" "$tmp_file"
 
   tinc foo stop
--- a/.builds/freebsd.yml
+++ b/.builds/freebsd.yml
@@ -1,6 +1,7 @@
 image: freebsd/13.x
 
 packages:
+  - coreutils
   - autotools
   - openssl
   - lzo2
--- a/.ci/tidy/run.sh
+++ b/.ci/tidy/run.sh
@@ -27,9 +27,7 @@ for path in $paths; do
 done
 
 if ! [ -f compile_commands.json ]; then
-  # Running compiledb directly on this doesn't work on FreeBSD for some reason.
-  make -j2 all extra
-  compiledb -n make check
+  compiledb make all extra
 fi
 
 echo >&2 "Running clang-tidy without $paths"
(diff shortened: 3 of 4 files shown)


hg:
UBSAN failure is triggered by hash table changes. I less sure about FreeBSD weirdness. I'll wait for hash table fixes, then rebase and investigate.

test/compression.test
```diff
@@ -106,11 +106,15 @@ for level in $levels; do
...
+    ip netns exec foo \
+      socat -u TCP4-LISTEN:$recv_port_foo,reuseaddr OPEN:"$tmp_file",creat || kill $$
+  ) &
+
+  (
+    ip netns exec bar \
+      socat -u OPEN:"$ref_file" TCP4:$ip_foo:$recv_port_foo,retry=30 || kill $$
+  ) &
 
   wait
```
gsliepen, line 119: This will still hang indefinitely if any of the socat commands don't terminate. Maybe it's better to just do something like:

```
cat > $tmp_file.sh << EOF
ip netns exec foo socat -u TCP4-LISTEN:$recv_port_foo,reuseaddr OPEN:"$tmp_file",creat &
ip netns exec bar socat -u OPEN:"$ref_file" TCP4:$ip_foo:$recv_port_foo,retry=30 &
wait
EOF

try_limit_time 60 sh -c $tmp_file.sh
```

gsliepen (commented):
> A bit of a rant: it's way too late for that, but after bashing my head against so many pointless differences in Unix utilities I am of the opinion that the test suite for cross-platform projects should be written in something like Python, or maybe even C.

Good point. I think it should just be written in C, no need to pull in yet more dependencies. Most of the external calls can just be done with `popen()`, and then you can just use `select()` with a timeout, or perhaps even simpler just use `alarm()`. Might need some work to get it Windows-proof, but in the end it could be much better than all the shell scripting hacks we are using now :)

test/compression.test
hg: Yes, this looks much better. Thank you.

It still could use the faster error detection (`|| kill $$`), but it's much uglier, so I removed it.

This all could have been written simpler (and it was), but of course there's old korn shell which handles background jobs unlike every other shell I've tested.

hg:
OK, the FreeBSD issue will have to be fixed by switching to GNU timeout. I don't want to further complicate the test suite. Its `timeout` behaves completely differently to GNU timeout and even other BSDs, and requires `--foreground` to cover our use case properly (and that flag is harmful everywhere else).

- https://github.com/hg/tinc/actions/runs/1131065003
- https://github.com/hg/tinc/actions/runs/1131080266
- https://github.com/hg/tinc/actions/runs/1131107368
- https://github.com/hg/tinc/actions/runs/1131126302
- https://builds.sr.ht/~reducer/job/566469
- https://builds.sr.ht/~reducer/job/566470

For some magical reason, compiledb recently stopped generating proper `compile_commands.json` on FreeBSD (see first sourcehut link), probably because of some compatibility weirdness with FreeBSD make. I pushed a drive-by fix for this (see the second link). It's still fine on [Linux](https://github.com/hg/tinc/runs/3330307420?check_suite_focus=true#step:6:188).

fangfufu:
Will this fix issue https://github.com/gsliepen/tinc/issues/306?

PR #318, opened 2021-08-15
simplify signal handling

splitice:
The splay tree for signal handling is unnecessary.


gsliepen:
Squashed & rebased as 290ca7a798604f33d30089477f0cddf7ba5eff9f. Thanks!

PR #323, opened 2021-08-16
epoll patch v2

splitice:
Version 2 of the epoll patch

### DIFF
--- a/src/event.c
+++ b/src/event.c
@@ -168,26 +168,21 @@ void io_set(io_t *io, int flags) {
 
 #ifndef HAVE_MINGW
 #ifdef HAVE_SYS_EPOLL_H
-	struct epoll_event ev;
-	ev.data.fd = io->fd;
-	ev.data.ptr = io;
-
 	epoll_ctl(epollset, EPOLL_CTL_DEL, io->fd, NULL);
-	io_tree.generation++;
 
-	if((flags & IO_READ) && (flags & IO_WRITE)) {
-		ev.events = EPOLLIN | EPOLLOUT;
+	struct epoll_event ev = {
+		.events = 0,
+		.data.ptr = io,
+	};
+
+	if(flags & IO_READ) {
+		ev.events |= EPOLLIN;
 	}
 
-	else if(flags & IO_READ) {
-		ev.events = EPOLLIN;
-	}
-
-	else if(flags & IO_WRITE) {
-		ev.events = EPOLLOUT;
-	}
-
-	else {
+	if(flags & IO_WRITE) {
+		ev.events |= EPOLLOUT;
+	} else if(ev.events == 0) {
+		io_tree.generation++;
 		return;
 	}
 
@@ -412,16 +407,9 @@ bool event_loop(void) {
 
 		int n = epoll_wait(epollset, events, EPOLL_MAX_EVENTS_PER_LOOP, (int)timeout);
 #else
-		int maxfds = 1;
+		int maxfds =  0;
 
 		if(io_tree.tail) {
-			/*
-			        Use the max FD number to detirmine the max size of events
-			        This abuses the fact that:
-			        a) tinc does not allocate new sockets during operation
-			        b) that tinc does not have many sockets
-			        c) that sockets are allocated from 0 on all systems and so max(fd) is a reasonably approximation of nfds
-			*/
 			io_t *last = io_tree.tail->data;
 			maxfds = last->fd + 1;
 		}
@@ -453,8 +441,16 @@ bool event_loop(void) {
 				io->cb(io->data, IO_WRITE);
 			}
 
-			if(events[i].events & EPOLLIN && curgen == io_tree.generation && io->flags & IO_READ) {
+			if(curgen != io_tree.generation) {
+				break;
+			}
+
+			if(events[i].events & EPOLLIN && io->flags & IO_READ) {
 				io->cb(io->data, IO_READ);
+			}
+
+			if(curgen != io_tree.generation) {
+				break;
 			}
 		}
 
--- a/src/event.c
+++ b/src/event.c
@@ -18,20 +18,43 @@
 */
 
 #include "system.h"
+#include "dropin.h"
+
+#ifdef HAVE_SYS_EPOLL_H
+#include <sys/epoll.h>
+#endif
+
 #include "event.h"
 #include "utils.h"
+#include "net.h"
 
 struct timeval now;
-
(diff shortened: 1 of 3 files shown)


src/event.c
```diff
@@ -356,22 +470,24 @@ bool event_loop(void) {
 			}
 
 			/*
-			   There are scenarios in which the callback will remove another io_t from the tree
```
splitice, line 359: make astyle is insistent on the level of indentation.

src/event.c
```diff
@@ -132,6 +167,35 @@ void io_set(io_t *io, int flags) {
...
 #ifndef HAVE_MINGW
+#ifdef HAVE_SYS_EPOLL_H
+	struct epoll_event ev;
+	ev.data.fd = io->fd;
+	ev.data.ptr = io;
+
+	epoll_ctl(epollset, EPOLL_CTL_DEL, io->fd, NULL);
+	io_tree.generation++;
+
+	if((flags & IO_READ) && (flags & IO_WRITE)) {
```
gsliepen, line 178: Can be made a bit cleaner by doing:

```
struct epoll_event ev = {
    .data.fd = io->fd;
    .data.ptr = io;
};
...
if(flags & IO_READ)
    ev.events |= EPOLLIN;

if(flags & IO_WRITE)
    ev.events |= EPOLLOUT;
```

src/event.c
```diff
@@ -132,6 +167,35 @@ void io_set(io_t *io, int flags) {
...
+	else if(flags & IO_READ) {
+		ev.events = EPOLLIN;
+	}
+
+	else if(flags & IO_WRITE) {
+		ev.events = EPOLLOUT;
+	}
+
+	else {
+		return;
```
gsliepen, line 191: It should be valid to call `epoll_ctl(EPOLL_CTL_ADD)` with `ev.events = 0`. Also, instead of doing `EPOLL_CTL_DEL` followed by `EPOLL_CTL_DEL`, why not use `EPOLL_CTL_MOD`?

src/event.c
```diff
@@ -314,23 +381,53 @@ bool event_loop(void) {
...
+#else
+		int maxfds = 1;
 
 		if(io_tree.tail) {
+			/*
+			        Use the max FD number to detirmine the max size of events
+			        This abuses the fact that:
+			        a) tinc does not allocate new sockets during operation
+			        b) that tinc does not have many sockets
+			        c) that sockets are allocated from 0 on all systems and so max(fd) is a reasonably approximation of nfds
```
gsliepen, line 423: Yeah there were some experiments randomizing fds that turned out to upset a lot of programs. But the comments make it sound we do something bad, but we don't: at this point io_tree.tail really should point to the event with the highest fd number, no higher fd should be set in the fdsets. Nothing is abused.

src/event.c
```diff
@@ -314,23 +381,53 @@ bool event_loop(void) {
...
+		long timeout = (tv->tv_sec * 1000) + (tv->tv_usec / 1000);
 
-		int fds = 0;
+		if(timeout > INT_MAX) {
+			timeout = INT_MAX;
+		}
+
+		int n = epoll_wait(epollset, events, EPOLL_MAX_EVENTS_PER_LOOP, (int)timeout);
+#else
+		int maxfds = 1;
```
gsliepen, line 415: Zero is fine if there really are no fds set.

src/event.c
```diff
@@ -346,6 +443,23 @@ bool event_loop(void) {
...
+			io_t *io = events[i].data.ptr;
+
+			if(events[i].events & EPOLLOUT && io->flags & IO_WRITE) {
+				io->cb(io->data, IO_WRITE);
+			}
+
+			if(events[i].events & EPOLLIN && curgen == io_tree.generation && io->flags & IO_READ) {
+				io->cb(io->data, IO_READ);
+			}
+		}
```
gsliepen, line 459: Yes, but we also need an `if(curgen != io_tre.generation) break;` at the end of the loop, because a reading data from one fd might cause data to be sent to another fd in the callback, and if that sending fails it might close the connection, and hence invalidate the `io_t` pointer of a subsequent event in the loop.

gsliepen (changes requested):
Apart from the issues I mentioned, it's looking nice and clean!

src/event.c
splitice: epoll mod won't add (ENOENT)
epoll add won't mod (EEXIST)

We don't know if the fd is being modified or added in the current interface so a DEL+ADD is required.

src/event.c
splitice: Biggest problems occur if tinc is linked with any shared library creating file descriptors for other purposes (e.g logging). Some comment explaining that you are using max(fd) to approximate nfds in the IO tree would be a good idea. But I'll remove it.

src/event.c
splitice: There would be gains in keeping fds permanently registered in general operation for both read at a given time and then being able to just EPOLL_CTL_MOD reliably to remove write as appropriate.

src/event.c
splitice: Done

src/event.c
splitice: Done

src/event.c
splitice: Done

src/event.c
splitice: Done

src/event.c
gsliepen: Ah, maybe you misunderstood the purpose of the first argument to `select()`? The parameter name in the manpage (and the one tinc used) is perhaps confusing, it's not the number of filedescriptors that we have or are interested in, it's rather the size of the fdsets in bits that we care about. If it's set to 10, it will check the first 10 bits in all the fdsets starting from the first one, so it will check fds 0 to 9.

src/event.c
gsliepen: I agree, it would be nice if tinc followed epoll requirements.

splitice:
Fyi from my end I am happy with the state post review.

src/event.c
gsliepen: Do you want to do this in this PR or want to do it in a separate one? I wouldn't mind the less efficient DEL+ADD for now to avoid having to make too many changes in one go.

splitice: Seperate for sure. Not certain the best way to do it yet.

gsliepen:
Squashed and rebased as b322126576c52ffbc9a91d818125fc461fa17d1b. Thanks!

PR #324, opened 2021-08-17
Leaks in `invitation.c` + minor Solaris fixes

hg:
- https://github.com/hg/tinc/actions/runs/1138466183
- https://github.com/hg/tinc/actions/runs/1138456174

Found in the process of making libgcrypt work. Cryptographic library failures were pushing `invitation.c` into previously untested error handling paths.

Fuzz testing could help with that, probably. I am not sure how to implement it, though: tincd networking code seems to require a lot of state being set up to function properly, so there has to be a network connection of some sort.

Running two separate tincd processes and sending fuzz data between them won't work — fuzzers monitor the very same process they're writing data into and expect it to fail.

I'm thinking of trying to create a separate thread, `connect()` to the same process, send data there, and see what happens.

---

For Solaris CI we could try this:

https://github.com/vmactions/solaris-vm

I'm getting the idea that if the code is not being regularly tested, then it quickly bitrots and probably does not work. Does it make sense to spend time on a pretty much dead OS?

### DIFF
--- a/src/invitation.c
+++ b/src/invitation.c
@@ -411,7 +411,7 @@ int cmd_invite(int argc, char *argv[]) {
 
 		if(!f) {
 			fprintf(stderr, "Could not write %s: %s\n", filename, strerror(errno));
-			free(key);
+			ecdsa_free(key);
 			return 1;
 		}
 
@@ -420,7 +420,7 @@ int cmd_invite(int argc, char *argv[]) {
 		if(!ecdsa_write_pem_private_key(key, f)) {
 			fprintf(stderr, "Could not write ECDSA private key\n");
 			fclose(f);
-			free(key);
+			ecdsa_free(key);
 			return 1;
 		}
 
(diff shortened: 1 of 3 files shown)


hg:
Hmm, I ran it locally on a Windows VM for a few dozen times before publishing this, and it was fine. I'll do more testing to determine if it's a spurious failure, or if there are new regressions.

hg:
After looping the test suite for two more hours, I think [this](https://github.com/hg/tinc/actions/runs/1138456174) was a one-off failure unrelated to these changes, unless I am missing something very obvious.

I also threw in a fix for [this](https://github.com/gsliepen/tinc/pull/326/checks?check_run_id=3349249334) for good measure (triggered by pure chance in #326 and caused by a recent refactoring). My bad.

https://github.com/hg/tinc/actions/runs/1138918466

gsliepen (approved):
I'm not too fond of Solaris myself, and I agree with the bit-rotting comment. Still, if we find something or people complain, it usually is not that much work to fix it, so why not keep the support for now. Setting up a VM for testing it in CI might be interesting, but on the other hand I would say that is low priority.

fangfufu:
Btw, @gsliepen, I noticed that you put "approved" on this pull request. What does that actually mean for us? I don't think @hg can merge the pull request himself, because he is not a "collaborator".

gsliepen:
> Btw, @gsliepen, I noticed that you put "approved" on this pull request. What does that actually mean for us? I don't think @hg can merge the pull request himself, because he is not a "collaborator".

I'll merge it later today!

PR #330, opened 2021-08-18
remove full subnet cache clear on graph update

splitice:
I wrote this, seemed like it would be a good idea. Thought I would add weight checks next, but I seem to have got my head in a loop in this graph code.

So I'm taking a break grom the graph code.

Effectively the goal would be to only clear from the hash cache if the incoming subnets have a lower weight than the currently cached item

This current version over clears, it doesnt do weight comparisons.


splitice:
Ergh, this is quite difficult to get right.

Issue #329, opened 2021-08-18
Create repositories for the nightly builds


fangfufu:
I think it would be great if we create binary repositories for tinc, since we are already creating nightly builds. This way it is easier to keep tinc updated. 

For Debian in particular, it seems to be possible to host the repository on Github itself: https://pmateusz.github.io/linux/2017/06/30/linux-secure-apt-repository.html

splitice:
Stable builds would be really useful too

PR #331, opened 2021-08-18
CI: improve sanitizer runs; update cross-compilation jobs to Debian bullseye

hg:
https://github.com/hg/tinc/actions/runs/1143479443

and a failing run to show it actually works:

https://github.com/hg/tinc/runs/3362255572

---

- sanitizers now do the full test run, as in every other job.
    - I would really like to merge this before libgcrypt, because there were a lot of sanitizer failures there. That's what prompted this change.
- run all test flavors even if one of them fails.
- move cross-compilation jobs to Debian Bullseye
    - big-endian MIPS is [no more](https://wiki.debian.org/SupportedArchitectures). I changed it to little-endian MIPS.
    - bullseye still ships the cross-compilation toolkit, but it's pretty useless for our purposes without pre-built libraries.

---

I think CI can finally be left alone after this. Maybe I'll add repositories as per #329, although I don't feel comfortable putting even more eggs in one basket (GitHub).

We could probably host them anywhere, and you could configure a subdomain to point it to GitHub for the time being, then easily move if Microsoft pulls a fast one.

To do this properly, we should use each distribution's own service:

- https://copr.fedorainfracloud.org
- https://launchpad.net/ubuntu/+ppas
- https://build.opensuse.org/

although TBH I don't have time to support this many.


.github/workflows/test.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       matrix:
         arch:
           - armhf
-          - mips
+          - mipsel
 
     container:
       image: debian:buster
```
gsliepen, line 22: The pull request mentions updating it to Debian bullseye, but it's still using buster?

fangfufu:
We could just use openSUSE Build Service. On https://build.opensuse.org/, it says 
>The openSUSE Build Service is the public instance of the Open Build Service (OBS) used for development of the openSUSE distribution and to offer packages from same source for Fedora, Debian, Ubuntu, SUSE Linux Enterprise and other distributions..

I get the official Wine Debian packages from openSUSE Build Service. 

I am more than happy to provide my own Kimsufi servers for hosting the repository, if required.

.github/workflows/test.yml
hg: That was pretty stupid of me. It was supposed to be a temporary change to find out the reason why qemu stopped working. Thanks.

- https://github.com/hg/tinc/actions/runs/1146226549
- https://github.com/hg/tinc/actions/runs/1146251417

hg:
>We should just use openSUSE Build Service

I think you're supposed to provide the source code and use their service to do the build (which is the right thing to do in any case). I'll probably look into doing it later. It doesn't seem to require any changes to CI.

gsliepen:
Cherry-picked as 53b239863ea0a44270d877945fa2014ec626ad01.

Issue #333, opened 2021-08-22
Sanitizer jobs failing


gsliepen:
@hg Commit 53b239863ea0a44270d877945fa2014ec626ad01, which seemed to pass testing itself, causes subsequent commits to fail the sanitizer job. I think the issue is that the tarball with the test logs were created as a job step before, and are now done by `.ci/test/run.sh`, and perhaps this trips up `.ci/sanitizers/run.sh` when it checks for any logs being present? Furthermore, the zip file created is called `tests_.zip`, the change in the "Upload test result" step probably has caused that.

hg:
Do you mean [this job](https://github.com/gsliepen/tinc/runs/3394652514?check_suite_focus=true#step:5:608) (and the other one with a similar failure)? I've tested that branch locally, UBSAN produced 137 log files, so it detects real issues with the code.

```
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior xoshiro.c:31:36 in
xoshiro.c:25:12: runtime error: left shift of 17758701374930733747 by 7 places cannot be represented in type 'uint64_t' (aka 'unsigned long')
    #0 0x55eae1475666 in rotl src/xoshiro.c:25:12
    #1 0x55eae147512d in xoshiro src/xoshiro.c:31:26
    #2 0x55eae1450a0a in prng src/./crypto.h:38:11
    #3 0x55eae144fcb2 in check_port src/tincctl.c:2130:28
    #4 0x55eae145f21f in cmd_init src/tincctl.c:2236:2
    #5 0x55eae145311a in cmd_shell src/tincctl.c:3232:15
    #6 0x55eae1450d65 in main src/tincctl.c:3309:10
    #7 0x7f0e92adeb24 in __libc_start_main (/usr/lib/libc.so.6+0x27b24)
    #8 0x55eae131b88d in _start (src/tinc+0x7588d)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior xoshiro.c:25:12 in
xoshiro.c:31:44: runtime error: unsigned integer overflow: 4164254924859070971 * 9 cannot be represented in type 'unsigned long'
    #0 0x55eae1475181 in xoshiro src/xoshiro.c:31:44
    #1 0x55eae1450a0a in prng src/./crypto.h:38:11
    #2 0x55eae144fcb2 in check_port src/tincctl.c:2130:28
    #3 0x55eae145f21f in cmd_init src/tincctl.c:2236:2
[...]

PR #334, opened 2021-08-23
CI: fix archive name for sanitizer results.

hg:
#333 

All sanitizer jobs were writing to the same name and overwriting each other.

---

Success:

https://github.com/hg/tinc/actions/runs/1157587095

---

The same fix with xoshiro changes applied on top:

https://github.com/hg/tinc/actions/runs/1157593127

UBSAN results (or see the link above):

https://github.com/hg/tinc/suites/3566773997/artifacts/85630898

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -143,7 +143,7 @@ jobs:
       - name: Upload test results
         uses: actions/upload-artifact@v2
         with:
-          name: tests_${{ env.ARTIFACT }}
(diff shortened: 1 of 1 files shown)


hg:
Some minor leaks remain in libgcrypt code, so it's not currently being sanitized. They are of constant size and have very short stack traces that point to initialization code in the library itself. I couldn't find any better solution than ignoring them (yet), I'll try to fix that until the end of the week.

PR #353, opened 2022-03-14
Silence remaining 'bugprone-narrowing-conversions'

hg:
It's not the most important thing in the world, but it's probably best to fix these warnings instead of turning off the check completely.

Note that it only shows up with `clang-tidy` 12 or newer.

Are you fine with JavaScript-ish casts `!!` or should I replace them with proper `(bool)foobar`?

### DIFF
--- a/src/connection.h
+++ b/src/connection.h
@@ -34,22 +34,25 @@
 #define OPTION_CLAMP_MSS        0x0008
 #define OPTION_VERSION(x) ((x) >> 24) /* Top 8 bits are for protocol minor version */
 
-typedef struct connection_status_t {
-	uint32_t pinged: 1;                 /* sent ping */
-	uint32_t unused_active: 1;
-	uint32_t connecting: 1;             /* 1 if we are waiting for a non-blocking connect() to finish */
-	uint32_t unused_termreq: 1;         /* the termination of this connection was requested */
-	uint32_t remove_unused: 1;          /* Set to 1 if you want this connection removed */
-	uint32_t timeout_unused: 1;         /* 1 if gotten timeout */
-	uint32_t encryptout: 1;             /* 1 if we can encrypt outgoing traffic */
-	uint32_t decryptin: 1;              /* 1 if we have to decrypt incoming traffic */
-	uint32_t mst: 1;                    /* 1 if this connection is part of a minimum spanning tree */
-	uint32_t control: 1;                /* 1 if this is a control connection */
-	uint32_t pcap: 1;                   /* 1 if this is a control connection requesting packet capture */
-	uint32_t log: 1;                    /* 1 if this is a control connection requesting log dump */
-	uint32_t invitation: 1;             /* 1 if this is an invitation */
-	uint32_t invitation_used: 1;        /* 1 if the invitation has been consumed */
-	uint32_t tarpit: 1;                 /* 1 if the connection should be added to the tarpit */
+typedef union connection_status_t {
+	struct {
+		bool pinged: 1;                 /* sent ping */
+		bool unused_active: 1;
+		bool connecting: 1;             /* 1 if we are waiting for a non-blocking connect() to finish */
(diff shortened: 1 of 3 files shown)


gsliepen:
Even better would be not to need to cast at all. Perhaps we can solve this more elegantly by moving to C11 (if Linus can, so can we). Then we can replace the bitfields, that need to have a well-defined width in order to serialize them, with `union`s containing unnamed `struct`s, like so:

```
typedef union connection_status_t {
    struct {
        bool pinged: 1;
        bool unused_active: 1;
        ...
    };
    uint32_t value;
};
```

This should avoid warnings from clang-tidy, and will make accessing it as a 32-bit int much cleaner as well.

hg:
Thanks, that's definitely much better.

I don't think there's a decent way to ask `autoconf` for anything newer than C99, even though it seems to have had support for C11 [since 2.70](https://lwn.net/Articles/839395/). We probably can't rely on magic that only been out for 1.5 years?

The check in `configure.ac` doesn't seem to be strictly necessary and gcc compiles the code even with `-std=c99 -Wpedantic` (loudly protesting along the way).

If you'd rather not have it, I'll remove it. FWIW, meson has proper support for setting the C standard.

lancethepants:
So I know that [tomato firmware](https://www.freshtomato.org/) which has tinc integration (with a nice gui I made) is using pre gcc-4.6 toolchains, which as far as I can tell, pre-dates c11. One of the toolchains is 4.2.4 (stupid old I know), so I'm not sure it'd even have `-std=c1x` experimental support . I know there are people using tinc with tomato. Trust me, I really hate being that guy that pipes up and is resistant to change. 
If it's any consolation I finally was forced to work with meson, and although I don't care for having to create cross files, it is workable.

lancethepants:
Now to be fair, the old toolchains tomato firmware use are only necessary for the kernel because of proprietary kernel modules.  They could use a much more recent toolchain for the majority of the userspace, or they could integrate a static binary built from a newer toolchain. So don't let me singlehandedly veto things if the majority think this is a move in the right direction.  tinc in tomato is my baby, so just trying to watch over it.

gsliepen:
@lancethepants I just checked, even GCC 4.1 supports unnamed `struct`s inside `union`s. That's one worry less!
@hg Nice. I would probably not bother to write an autoconf check, recent versions of autoconf seem to already automatically emit checks for C11 support.

src/connection.h
```diff
@@ -34,24 +34,29 @@
...
+		bool pcap: 1;                   /* 1 if this is a control connection requesting packet capture */
+		bool log: 1;                    /* 1 if this is a control connection requesting log dump */
+		bool invitation: 1;             /* 1 if this is an invitation */
+		bool invitation_used: 1;        /* 1 if the invitation has been consumed */
+		bool tarpit: 1;                 /* 1 if the connection should be added to the tarpit */
+	};
+	uint32_t value;
 } connection_status_t;
 
+_Static_assert(sizeof(connection_status_t) == sizeof(uint32_t), "Wrong size for connection_status_t");
```
gsliepen, line 58: I would remove this check so the code can still be compiled with older GCC versions.

configure.ac
```diff
@@ -239,6 +238,13 @@ AC_ARG_ENABLE(jumbograms,
   ]
 )
 
+dnl Make sure that C11 standard is supported
+AX_CHECK_COMPILE_FLAG([-std=c11], [
+  CFLAGS="$CFLAGS -std=c11"
+], [
+  AC_MSG_ERROR("The C11 standard is not supported by your C compiler.")
+])
+
```
gsliepen, line 247: I would omit this whole commit.

hg:
Thanks for the review.

[Here's a test](https://godbolt.org/z/s8TfvKGc9) under gcc 4.1.2 just to be sure (which is 15 years old at this point).

It compiles even with `-std=gnu89` (or without specifying any standard).

PR #354, opened 2022-03-16
Add support for OpenSSL 3.0+

### LINKED ISSUE #347: Add clean support for OpenSSL 3
Currently tinc 1.0 and 1.1 compile and link with OpenSSL 3, but several warnings are generated about the use of depreacated functions. Ensure we use functions don't generate warnings.

hg:
This is an attempt to close #347.

Since this touches cryptography you might prefer to do the work yourself. I'm okay with that; this PR could be used as a starting point.

I tested it on one of the nodes in my 20-something node network. Seems to work fine.

Here's the official [migration guide](https://www.openssl.org/docs/man3.0/man7/migration_guide.html). It's not of much help, the information on how to use the new APIs has to be picked piecemeal in OpenSSL sources, manpages, and examples.

### DIFF
--- a/src/openssl/digest.c
+++ b/src/openssl/digest.c
@@ -22,9 +22,14 @@
 #include <openssl/err.h>
 #include <openssl/hmac.h>
 
+#if OPENSSL_VERSION_MAJOR >= 3
+#include <openssl/core_names.h>
+#endif
+
 #include "digest.h"
 #include "../digest.h"
 #include "../logger.h"
+#include "log.h"
 
 static void digest_open(digest_t *digest, const EVP_MD *evp_md, size_t maclength) {
 	digest->digest = evp_md;
@@ -63,13 +68,51 @@ bool digest_open_by_nid(digest_t *digest, int nid, size_t maclength) {
 }
 
 bool digest_set_key(digest_t *digest, const void *key, size_t len) {
+#if OPENSSL_VERSION_MAJOR < 3
 	digest->hmac_ctx = HMAC_CTX_new();
-	HMAC_Init_ex(digest->hmac_ctx, key, (int)len, digest->digest, NULL);
 
 	if(!digest->hmac_ctx) {
 		abort();
 	}
 
+	HMAC_Init_ex(digest->hmac_ctx, key, (int)len, digest->digest, NULL);
+#else
+	EVP_MAC *mac = EVP_MAC_fetch(NULL, OSSL_MAC_NAME_HMAC, NULL);
+
+	if(!mac) {
+		openssl_err("fetch MAC");
+		return false;
+	}
+
+	digest->hmac_ctx = EVP_MAC_CTX_new(mac);
+	EVP_MAC_free(mac);
+
+	if(!digest->hmac_ctx) {
+		openssl_err("create MAC context");
+		return false;
+	}
+
+	const char *hmac_algo = EVP_MD_get0_name(digest->digest);
+
+	if(!hmac_algo) {
+		openssl_err("get HMAC algorithm name");
+		return false;
+	}
+
+	// The casts are okay, the parameters are not going to change. For example, see:
+	// https://github.com/openssl/openssl/blob/31b7f23d2f958491d46c8a8e61c2b77b1b546f3e/crypto/ec/ecdh_kdf.c#L37-L38
+	const OSSL_PARAM params[] = {
+		OSSL_PARAM_octet_string(OSSL_MAC_PARAM_KEY, (void *)key, len),
+		OSSL_PARAM_utf8_string(OSSL_MAC_PARAM_DIGEST, (void *)hmac_algo, 0),
+		OSSL_PARAM_END,
+	};
+
+	if(!EVP_MAC_init(digest->hmac_ctx, NULL, 0, params)) {
+		openssl_err("set MAC context params");
+		return false;
+	}
+
+#endif
 	return true;
 }
 
@@ -83,7 +126,11 @@ void digest_close(digest_t *digest) {
 	}
 
 	if(digest->hmac_ctx) {
+#if OPENSSL_VERSION_MAJOR < 3
 		HMAC_CTX_free(digest->hmac_ctx);
+#else
+		EVP_MAC_CTX_free(digest->hmac_ctx);
+#endif
 	}
 
 	memset(digest, 0, sizeof(*digest));
@@ -94,10 +141,24 @@ bool digest_create(digest_t *digest, const void *indata, size_t inlen, void *out
 	unsigned char tmpdata[len];
 
 	if(digest->hmac_ctx) {
-		if(!HMAC_Init_ex(digest->hmac_ctx, NULL, 0, NULL, NULL)
-		                || !HMAC_Update(digest->hmac_ctx, indata, inlen)
-		                || !HMAC_Final(digest->hmac_ctx, tmpdata, NULL)) {
-			logger(DEBUG_ALWAYS, LOG_DEBUG, "Error creating digest: %s", ERR_error_string(ERR_get_error(), NULL));
+		bool ok;
+
+#if OPENSSL_VERSION_MAJOR < 3
+		ok = HMAC_Init_ex(digest->hmac_ctx, NULL, 0, NULL, NULL)
+		     && HMAC_Update(digest->hmac_ctx, indata, inlen)
+		     && HMAC_Final(digest->hmac_ctx, tmpdata, NULL);
+#else
+		EVP_MAC_CTX *mac_ctx = EVP_MAC_CTX_dup(digest->hmac_ctx);
+
+		ok = mac_ctx
+		     && EVP_MAC_update(mac_ctx, indata, inlen)
+		     && EVP_MAC_final(mac_ctx, tmpdata, NULL, sizeof(tmpdata));
+
+		EVP_MAC_CTX_free(mac_ctx);
+#endif
+
+		if(!ok) {
+			openssl_err("create HMAC");
 			return false;
 		}
 	} else {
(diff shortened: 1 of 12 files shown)


hg:
The CI run is [here](https://github.com/hg/tinc/actions/runs/1993898685). It seems GitHub is not starting another one if the original is successful.

src/openssl/digest.c
```diff
@@ -94,10 +143,27 @@ bool digest_create(digest_t *digest, const void *indata, size_t inlen, void *out
...
+		// The cast is okay, the parameter is not going change. For example, see:
+		// https://github.com/openssl/openssl/blob/31b7f23d2f958491d46c8a8e61c2b77b1b546f3e/crypto/ec/ecdh_kdf.c#L37-L38
+		const OSSL_PARAM params[] = {
+			OSSL_PARAM_utf8_string(OSSL_STORE_PARAM_DIGEST, (char *)digest->hmac_algo, 0),
+			OSSL_PARAM_END,
+		};
+
+		ok = EVP_MAC_init(digest->hmac_ctx, digest->hmac_key, digest->hmac_key_len, params)
+		     && EVP_MAC_update(digest->hmac_ctx, indata, inlen)
+		     && EVP_MAC_final(digest->hmac_ctx, tmpdata, NULL, sizeof(tmpdata));
```
gsliepen, line 162: Initializing a HMAC context is slow. That's why we want to call it with NULL parameters, otherwise you'd just use the one-shot functions, like `EVP_Q_mac()`. It looks like you should still be able to do pass NULL key and params if you set the key beforehand using `EVP_MAC_CTX_set_params()`, although I do wonder if it still caches all the state derived from the key...

src/openssl/digest.h
```diff
@@ -25,7 +25,14 @@
 
 struct digest {
 	const EVP_MD *digest;
+#if OPENSSL_VERSION_MAJOR < 3
 	HMAC_CTX *hmac_ctx;
+#else
+	EVP_MAC_CTX *hmac_ctx;
+	void *hmac_key;
+	size_t hmac_key_len;
+	const char *hmac_algo;
```
gsliepen, line 34: These things can be stored inside `hmac_ctx`, so no need to keep copies (I hope)?

gsliepen (commented):
> Since this touches cryptography you might prefer to do the work yourself. I'm okay with that; this PR could be used as a starting point.

I'm totally fine with someone else doing the work and me just having to review it. That also means at least two pairs of eyeballs have seen the code. I'm quite disappointed with the new OpenSSL 3.0 API, it doesn't make things any easier, it just replaces some old warts with new ones.

hg:
Context reuse doesn't seem to be [supported](https://github.com/openssl/openssl/issues/17811) by the new API. Looks like an accidental omission.

I [tried](https://github.com/hg/tinc/commit/a77544d027601ec67c5476ddef6b3c6e9415f470) to do it without much success. The tests pass, but the binary fails to connect to other nodes running the original version of tinc:

```
Failed to decrypt and verify record from xxx (xx.yy.zz.qq port 655)
Failed to decrypt and verify record from yyy (xx.yy.zz.qq port 655)
…
```

So here's another solution. The context is initialized once and then duplicated each time it needs to be used. Seems to work fine.

I don't have a fast enough network to check everything properly, so here's a couple of dumb benchmarks running on a single machine:

# reinit every time

```
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  5]   0.00-1.00   sec  35.0 MBytes   294 Mbits/sec   27    392 KBytes
[  5]   1.00-2.00   sec  32.5 MBytes   272 Mbits/sec    0    451 KBytes
[  5]   2.00-3.00   sec  33.8 MBytes   283 Mbits/sec    0    506 KBytes
[  5]   3.00-4.00   sec  33.8 MBytes   283 Mbits/sec    0    556 KBytes
[  5]   4.00-5.00   sec  35.0 MBytes   294 Mbits/sec    0    601 KBytes
[  5]   5.00-6.00   sec  45.0 MBytes   377 Mbits/sec    0    656 KBytes
[  5]   6.00-7.00   sec  45.0 MBytes   378 Mbits/sec    2    508 KBytes
[  5]   7.00-8.00   sec  45.0 MBytes   377 Mbits/sec    0    577 KBytes
[  5]   8.00-9.00   sec  43.8 MBytes   367 Mbits/sec    0    626 KBytes
[  5]   9.00-10.00  sec  45.0 MBytes   377 Mbits/sec    0    675 KBytes
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec   394 MBytes   330 Mbits/sec   29             sender
[  5]   0.00-10.01  sec   392 MBytes   328 Mbits/sec                  receiver
```

# duplicate

```
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  5]   0.00-1.00   sec  42.5 MBytes   356 Mbits/sec   96    687 KBytes
[  5]   1.00-2.00   sec  40.0 MBytes   336 Mbits/sec    6    537 KBytes
[  5]   2.00-3.00   sec  40.0 MBytes   336 Mbits/sec    0    591 KBytes
[  5]   3.00-4.00   sec  40.0 MBytes   336 Mbits/sec    0    643 KBytes
[  5]   4.00-5.00   sec  41.2 MBytes   346 Mbits/sec    6    474 KBytes
[  5]   5.00-6.00   sec  51.2 MBytes   430 Mbits/sec  139    556 KBytes
[  5]   6.00-7.00   sec  50.0 MBytes   419 Mbits/sec    9    440 KBytes
[  5]   7.00-8.00   sec  50.0 MBytes   419 Mbits/sec    0    522 KBytes
[  5]   8.00-9.00   sec  51.2 MBytes   430 Mbits/sec    0    591 KBytes
[  5]   9.00-10.00  sec  50.0 MBytes   419 Mbits/sec    0    653 KBytes
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec   456 MBytes   383 Mbits/sec  256             sender
[  5]   0.00-10.01  sec   453 MBytes   380 Mbits/sec                  receiver
```

# OpenSSL 1.1

```
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  5]   0.00-1.00   sec  40.6 MBytes   341 Mbits/sec   19    561 KBytes
[  5]   1.00-2.00   sec  38.8 MBytes   325 Mbits/sec    0    628 KBytes
[  5]   2.00-3.00   sec  38.8 MBytes   325 Mbits/sec    0    673 KBytes
[  5]   3.00-4.00   sec  37.5 MBytes   315 Mbits/sec    1    508 KBytes
[  5]   4.00-5.00   sec  40.0 MBytes   336 Mbits/sec    0    563 KBytes
[  5]   5.00-6.00   sec  50.0 MBytes   419 Mbits/sec    0    628 KBytes
[  5]   6.00-7.00   sec  48.8 MBytes   409 Mbits/sec    0    686 KBytes
[  5]   7.00-8.00   sec  50.0 MBytes   419 Mbits/sec    3    550 KBytes
[  5]   8.00-9.00   sec  50.0 MBytes   419 Mbits/sec    0    611 KBytes
[  5]   9.00-10.00  sec  50.0 MBytes   419 Mbits/sec    0    670 KBytes
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec   444 MBytes   373 Mbits/sec   23             sender
[  5]   0.00-10.01  sec   442 MBytes   370 Mbits/sec                  receiver
```

hg:
Cleaned up some of the copypasta. Not sure about `rsa_public_encrypt()` / `rsa_private_decrypt()`. They're exactly the same, except for the names of two functions they call. Should I extract common logic to this:

```c
static bool rsa_encrypt_decrypt(rsa_t *rsa, const void *in, size_t len, void *out,
                                int (init)(EVP_PKEY_CTX *ctx),
                                int (process)(EVP_PKEY_CTX *ctx,
                                                unsigned char *out, size_t *outlen,
                                                const unsigned char *in, size_t inlen)
                               ) {
	EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new(rsa, NULL);

	if(ctx) {
		size_t outlen = len;

		bool ok = init(ctx) > 0
		          && EVP_PKEY_CTX_set_rsa_padding(ctx, RSA_NO_PADDING) > 0
		          && process(ctx, out, &outlen, in, len) > 0
		          && outlen == len;

		EVP_PKEY_CTX_free(ctx);

		if(ok) {
			return true;
		}
	}
}

// …
return rsa_encrypt_decrypt(rsa, in, len, out, EVP_PKEY_decrypt_init, EVP_PKEY_decrypt);

// …
return rsa_encrypt_decrypt(rsa, in, len, out, EVP_PKEY_encrypt_init, EVP_PKEY_encrypt);
```

or is it too much for your taste?

Same thing with `cipher_encrypt()` / `cipher_decrypt()`.

I'm more used to working in higher-level languages where you wouldn't think twice about extracting common pieces and using such redirection.

gsliepen:
> So here's another solution. The context is initialized once and then duplicated each time it needs to be used. Seems to work fine.

Well, it seems it has comparable performance to the old solution, so it's fine with me. Nice benchmarks!

> Cleaned up some of the copypasta. Not sure about rsa_public_encrypt() / rsa_private_decrypt(). They're exactly the same, except for the names of two functions they call. Should I extract common logic to this: [...] or is it too much for your taste?

I'm fine with that, as long as code outside the `openssl/` directory does not have to see any OpenSSL details. So if you make `rsa_public_encrypt()` and `rsa_public_decrypt()` wrapper functions that call `rsa_encrypt_decrypt()`, that's fine.

> I'm more used to working in higher-level languages where you wouldn't think twice about extracting common pieces and using such redirection.

Yes, if I had a lot more time I would not hesitate spending it on rewriting tinc in C++ or Rust.

hg:
Thanks so much for putting your time in these reviews.

I cleaned up the remaining copypasta in `src/openssl`. There's not much difference in SLOC, but at least less logic is duplicated.

> rewriting tinc in C++ or Rust

I thought about that, but didn't even dare to bring it up because it sounds like a sure way to lose at least some long-time users. If devices still rely on GCC 4.2, the Rust cross-compilation story there is probably nonexistent.

Looking at the [official docs](https://doc.rust-lang.org/nightly/rustc/platform-support.html), the majority of targets that are important to tinc have these clauses:

> not guaranteed to produce a working build
> tier 2 targets often work to quite a good degree

It might just be Rust devs being very cautious with their wording, but it doesn't inspire much confidence.

C++ may actually be workable. GCC has (almost) full C++11 support [since 4.8.1](https://en.cppreference.com/w/cpp/compiler_support/11) (which is 9 years old), and the language lends itself to porting things slowly instead of doing full rewrites.

PR #355, opened 2022-03-18
Change the build system to meson?

hg:
The PR is here mostly to have a better place for discussing the possible move (and so I don't lose the work again).

See history in #286 (which isn't really about meson).

CI is barely working, I'd prefer to finish it last if the change is agreed upon as it's a pain to maintain otherwise.

---

Running `meson setup` on freshly cloned source tree is pretty fast:

```
Fri, 18 Mar 2022 18:47:02 GMT # Running test flavor default
Fri, 18 Mar 2022 18:47:02 GMT ################################################################################
…
Fri, 18 Mar 2022 18:47:05 GMT ninja: entering directory 'build'
Fri, 18 Mar 2022 18:47:05 GMT [1/113] Compiling C object test/splice.p/splice.c.o
```

vs autotools:

```
Fri, 18 Mar 2022 14:49:08 GMT # Running test flavor default
Fri, 18 Mar 2022 14:49:08 GMT ################################################################################
…
Fri, 18 Mar 2022 14:49:19 GMT make[1]: Entering directory '/__w/tinc/tinc'
```

All in all it seems to save 20-50 seconds on each test run ([autotools](https://github.com/hg/tinc/actions/runs/2004774852), [meson](https://github.com/hg/tinc/actions/runs/2005869521)).

---

Cross-compilation from Debian amd64 to mipsel and armhf requires these cross files:

```ini
[binaries]
c = 'arm-linux-gnueabihf-gcc'
pkgconfig = 'arm-linux-gnueabihf-pkg-config'
```

```ini
[binaries]
c = 'mipsel-linux-gnu-gcc'
pkgconfig = 'mipsel-linux-gnu-pkg-config'
```

or setting two environment variables (`CC` and `PKG_CONFIG`). Everything else is detected automatically.

So basically we have

    $ CC=arm-linux-gnueabihf-gcc PKG_CONFIG=arm-linux-gnueabihf-pkg-config meson setup build --cross-file /dev/null

or

    $ meson setup build --cross-file .ci/cross/armhf

vs current

    $ ./configure --whatever --host=arm-linux-gnueabihf-pkg-config

---

I am not sure how to go about trying to cross-compile for Windows. mingw ships only a couple of prebuilt libraries. @gsliepen do you build everything from source?

I believe it requires a larger file, although it too can be written once and stored in the repository:

```ini
[binaries]
c = 'x86_64-w64-mingw32-gcc'
pkgconfig = 'x86_64-w64-mingw32-pkg-config'

[host_machine]
system = 'windows'
endian = 'little'
cpu = 'x86_64'
cpu_family = 'x86'
```

If your compiler is named differently, you can use the same file and override just the compiler with `CC=xxx`.


meson.build
```diff
@@ -0,0 +1,43 @@
...
+prefix = get_option('prefix')
+bindir = prefix / get_option('bindir')
+mandir = prefix / get_option('mandir')
+libdir = prefix / get_option('libdir')
+datadir = prefix / get_option('datadir')
+infodir = prefix / get_option('infodir')
+sbindir = prefix / get_option('sbindir')
+localedir = prefix / get_option('localedir')
+sysconfdir = prefix / get_option('sysconfdir')
+localstatedir = prefix / get_option('localstatedir')
```
gsliepen, line 28: You can't do this, some directories will be absolute and not live inside prefix, by default that's `syconfdir`, `localstatedir` and `sharedstatedir`. I think you should use [`join_path()`](https://mesonbuild.com/Reference-manual_functions.html#join_paths) here, like so:

```
foodir = join_path(prefix, get_option('foodir')
```

src/meson.build
```diff
@@ -0,0 +1,424 @@
...
+    cdata.set('HAVE_READLINE', 1)
+    tinc_dep += dep_readline
+  endif
+endif
+
+dep_zlib = dependency('zlib', required: get_option('zlib'))
+if dep_zlib.found() and cc.has_function('compress2', dependencies: dep_zlib)
+  cdata.set('HAVE_ZLIB', 1)
+  tincd_dep += dep_zlib
+endif
```
gsliepen, line 305: Hm, I wonder if this is the proper way to do it. First of all, I think you can just unconditionally add the dependency object to `tincd_dep`; if it's not required then it should be a no-op. Also, meson is quite a bit smarter than autotools when it comes to checking dependencies; I don't think we need to check for `compress2()`. At most we should specify a minimum version for the dependency.

src/openssl/meson.build
```diff
@@ -0,0 +1,46 @@
...
+
+if not dep_openssl.found()
+  dep_openssl = dependency('openssl')
+endif
+
+foreach f : ['HMAC_CTX_new', 'BN_GENCB_new', 'RSA_set0_key']
+  if cc.has_function(f, dependencies: dep_openssl)
+    cdata.set('HAVE_' + f.to_upper(), 1)
+  endif
+endforeach
```
gsliepen, line 11: Is there a way to avoid this?

src/solaris/meson.build
```diff
@@ -0,0 +1,9 @@
+dep_socket = cc.find_library('socket')
+if not cc.has_function('connect', dependencies: dep_socket)
+  error('required function connect missing')
+endif
```
gsliepen, line 4: We should not check for this. I think it was necessary with autotools to properly detect the presence of the library.

test/security.test
```diff
@@ -101,7 +102,7 @@ pid=$!
 require_nodes foo 1
 require_nodes bar 1
 
-kill $pid
+kill $pid || true
```
gsliepen, line 105: Why is `|| true` necessary? That doesn't seem meson-related?

test/meson.build
```diff
@@ -0,0 +1,47 @@
...
+env.set('TINC_PATH', tinc_exe.full_path())
+env.set('TINCD_PATH', tincd_exe.full_path())
+env.set('SPTPS_TEST_PATH', sptps_test_exe.full_path())
+env.set('SPTPS_KEYPAIR_PATH', sptps_keypair_exe.full_path())
+env.set('SPLICE_PATH', splice_exe.full_path())
+env.set('TESTLIB_PATH', src_root / 'test' / 'testlib.sh')
+
+foreach test_name : tests
+  target = find_program(test_name, native: true)
+  test(test_name, target, timeout: 600, env: env, is_parallel: false)
```
gsliepen, line 45: Oh, a built-in test timeout, how nice!

gsliepen:
> I am not sure how to go about trying to cross-compile for Windows. mingw ships only a couple of prebuilt libraries. @gsliepen do you build everything from source?

I managed with a cross file similar to the one you made for MinGW. I do build the dependencies from source, it's just making sure autotools/meson can find them.

src/meson.build
hg: Even the version check is probably pretty [pointless](https://zlib.net/ChangeLog.txt):

    Changes in 1.0.8 (27 Jan 1998)
    …
    - added compress2() to allow setting the compression level

meson.build
hg: I believe it's the same thing?

> (since 0.49.0) Using the / operator on strings is equivalent to calling join_paths().

> If any one of the individual segments is an absolute path, all segments before it are dropped.

I was working from [this](https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/b868ec36c5d16c3d89b1c2005a9f0b9b24a7d415/meson.build#L31-41) and [this](https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/blob/852c15954ee2c02b8a82cf98f2996adedf6ca0b8/meson.build#L60-73).

.ci/test/run.sh
```diff
@@ -69,10 +58,11 @@ MINGW*)
...
 Darwin)
-  nproc() { sysctl -n hw.ncpu; }
-  gcrypt=$(brew --prefix libgcrypt)
-  openssl=$(brew --prefix openssl)
-  export CPPFLAGS="${CPPFLAGS:-} -I/usr/local/include -I$gcrypt/include -I$openssl/include -I$gcrypt/include"
+  sudo() { "$@"; }
+  #  nproc() { sysctl -n hw.ncpu; }
+  #  gcrypt=$(brew --prefix libgcrypt)
+  #  openssl=$(brew --prefix openssl)
+  #  export CPPFLAGS="${CPPFLAGS:-} -I/usr/local/include -I$gcrypt/include -I$openssl/include -I$gcrypt/include"
```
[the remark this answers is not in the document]
hg: Ah, sorry for the mess. I was not sure this whole idea won't be rejected at first sight, so all changes related to CI were in pretty bad shape, as mentioned in the original comment.

Should be fixed now.

hg:
Windows is a pain, as usual. I look into fixing it next.

meson.build
gsliepen: Ah, I missed that. Then it's fine!

src/meson.build
```diff
@@ -0,0 +1,425 @@
...
+  sptps_test_src += src
+  sptps_keypair_src += src
+endif
+
+# No pkg-config files on MinGW
+opt_miniupnpc = get_option('miniupnpc')
+dep_miniupnpc = dependency('miniupnpc', required: opt_miniupnpc, static: win_static)
+if not dep_miniupnpc.found()
+  dep_miniupnpc = cc.find_library('miniupnpc', required: opt_miniupnpc, static: win_static)
+endif
```
gsliepen, line 265: If you pass `-D miniupnpc=enabled`, then if `dependency()` doesn't find the MiniUPnP library, it will immediately stop with an error, it will not try the `cc.find_library()` method. Also, if you pass `-D miniupnpc=disabled`, it will do both checks. So perhaps it should be rewritten to:

```
if opt_miniupnpc.allowed():
    dep_miniupnpc = dependency('miniupnpc', required: false, static: win_static)
    if not dep.miniupnpc.found()
        dep_miniupnpc = cc.find_library('miniupnpc', required: opt_miniupnpc, static: win_static)
    endif
    if dep.miniupnpc.found()
         # Add dependency to tincd_dep
         ...
    endif
endif
```

src/meson.build
```diff
@@ -0,0 +1,425 @@
...
+  tincd_src += src
+  sptps_test_src += src
+  sptps_keypair_src += src
+endif
+
+# No pkg-config files on MinGW
+opt_miniupnpc = get_option('miniupnpc')
+dep_miniupnpc = dependency('miniupnpc', required: opt_miniupnpc, static: win_static)
+if not dep_miniupnpc.found()
+  dep_miniupnpc = cc.find_library('miniupnpc', required: opt_miniupnpc, static: win_static)
```
gsliepen, line 264: My meson is complaining that `static: win_static` inside `find_library()` is a version 0.51 feature, but you declared the minimum meson version to be 0.49.

hg: Apologies, I didn't intend to bug you with what was essentially work in progress.

The functionality is more or less finished, all features finally work and [packages](https://github.com/hg/tinc/releases/tag/latest) build without errors.

I'll try to clean up the configs next. Should I remove autotools files, or do we keep both for now?

meson.build
```diff
@@ -0,0 +1,87 @@
...
+endif
+
+cc_flags = cc.get_supported_arguments(cc_flags)
+ld_flags = cc.get_supported_link_arguments(ld_flags)
+
+add_project_arguments(cc_flags, language: 'c')
+add_project_link_arguments(ld_flags, language: 'c')
+
+build_root = meson.build_root()
+src_root = meson.source_root()
```
eli-schwartz, line 61: https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonbuild_root

Deprecated in 0.56

> Try using [meson.current_build_dir()](https://mesonbuild.com/Reference-manual_builtin_meson.html#mesoncurrent_build_dir) or [meson.project_build_root()](https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonproject_build_root). In the rare cases where the root of the main project is needed, use [meson.global_build_root()](https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonglobal_build_root) that has the same behaviour but with a more explicit name.

src/meson.build
```diff
@@ -0,0 +1,428 @@
...
+  if dep_lz4.found()
+    tincd_dep += dep_lz4
+    cdata.set('HAVE_LZ4', 1)
+    cdata.set('LZ4_H', '<lz4.h>')
+  elif not opt_lz4.auto()
+    tincd_src += 'lib/lz4/lz4.c'
+    cdata.set('HAVE_LZ4', 1)
+    cdata.set('LZ4_H', '"lib/lz4/lz4.h"')
+  endif
+endif
```
eli-schwartz, line 336: Upstream lz4 has a meson.build, which is wrapped in the Meson WrapDB.

```
mkdir -p subprojects
meson wrap install lz4
```

And then `dependency('lz4', required: get_option('lz4'))` can automatically fall back to the subproject, download it on demand (if wrap-mode doesn't specify `nodownload`/`nofallback`) and build it as a private copy.

src/meson.build
gsliepen: Since you already broke the autotools, perhaps it's best to remove it.

src/openssl/meson.build
```diff
@@ -0,0 +1,39 @@
+dep_openssl = dependency('openssl11', required: false, static: static)
+if not dep_openssl.found()
+  dep_openssl = dependency('openssl', static: static)
```
eli-schwartz, line 3: Should this have a versioned requirement to ensure it only detects openssl 1.1.x?

gsliepen: Good point, it should as we already dropped support for older versions.

eli-schwartz: (I am honestly not sure why any distros are patching out that pkg-config file to have the wrong name.)

src/meson.build
```diff
@@ -0,0 +1,428 @@
...
+
+check_functions = [
+  'asprintf',
+  'daemon',
+  'devname',
+  'fchmod',
+  'fdevname',
+  'fork',
+  'gettimeofday',
+  'mlockall',
```
gsliepen, line 64: The check for `mlockall()` is failing because `<sys/mman.h>` is not `#include`d while running the test. It has to be added to `src/have.h`. Alternatively, according to the manpage it seems that:

> On POSIX systems on which `mlockall()` and `munlockall()` are available, `_POSIX_MEMLOCK` is defined in `<unistd.h>` to a value greater than 0.  (See also sysconf(3).)

So we could use that instead of checking the function here.

hg:
I extracted some shared sources into separate static libraries, since it saves about a third of CPU time:

    $ time ninja -C build
    real	0m6.716s
    user	0m21.727s
    sys	0m2.651s

and after:

    $ time ninja -C build
    real	0m4.459s
    user	0m14.399s
    sys	0m1.838s

although the dependency tree is now less obvious.

If you'd rather keep it simple, I'll revert the change.

---

I'd like to restore Android build instructions in a separate PR a few days from now, as I currently don't have the means to download Android SDK (which is pretty large).

src/meson.build
eli-schwartz: Great, happy to help.

There are other wraps in the WrapDB which you may want to take advantage of too. :)

src/meson.build
```diff
@@ -340,6 +340,18 @@ if opt_crypto == 'nolegacy'
...
+    if dep_openssl.found()
+      break
+    endif
+  endforeach
+  if not dep_openssl.found()
+    dep_openssl = dependency('',
+                             required: opt_crypto == 'openssl' or opt_tests.enabled(),
+                             static: static,
+                             fallback: ['openssl', 'openssl_dep'])
+  endif
```
eli-schwartz, line 354: This seems like a lot of fairly confusing conditionals. Note that in Meson 0.60.0 you can specify multiple names for the same dependency, and in 0.62.0 (shortly to be released) looking up the name "openssl" specifically, will automatically try a "system" class fallback that probes for the library names and the header with the version.

The former requires bumping your minimum supported version of Meson. The latter is fully backwards compatible as long as you are okay telling people "your weird system will only find openssl if you upgrade Meson, but people with normal systems that have normal pkg-config files will have it work even on older versions of Meson".

hg: Maybe put a TODO for now? I wanted to use the multi-name feature, but it's probably best to support whatever is available in most popular Linux distributions (0.56 in Debian stable, 0.55 in RHEL 8, 0.53 in Ubuntu LTS, etc) without forcing users to install meson manually. Package maintainers in particular are not likely to be happy about this.

Let's see what @gsliepen thinks.

eli-schwartz: On the one hand:
- distro maintainers for stable (LTS) distros with older versions of Meson aren't going to package a new version of any software for any reason whatsoever, and thus don't care what version of Meson you use
- regular users on stable (LTS) distros can `pip install` a new version of Meson pretty easily

On the other hand, it might be desirable for users to be able to just have it work using the distro version of Meson.

On the other other hand, depending on which LTS distro is under discussion they may have an end of life version of Python. Meson 0.62.0 drops support for python 3.6.

On the other other other hand, the BSDs are notorious for bundling software in the base system, then refusing to provide pkg-config files for them at all. You have no options for detecting openssl on FreeBSD whatsoever, other than upgrading Meson to 0.62.0 (in fact, FreeBSD is ***exactly*** the reason I added the custom lookup) or falling back to `cc.find_library()` pain. Admittedly that is something you'd have to do with autotools too...

On the other other other other hand, you can just tell people "okay we give up, use the wrapped subproject for OpenSSL". :D That will work for about a dozen versions back.

hg: OK, I apologize for wasting your time.

This particular snippet was written way back when `sptps_keypair` and `sptps_speed` required OpenSSL whether you're using it for the two main programs or not.

It was fixed in 77cd819058de43a5fcea54300dde50e03088c318, but I completely forgot about it here.

Having two build systems in parallel is probably not a good idea.

eli-schwartz: Ah, okay, at least then you don't need to fallback to `dependency('', ...)`

:)

eli-schwartz: The new approach for a single use case is pretty clean, and probably the best that can be done without "multiple names" support. At that point it's really just a style difference.

hg:
I tried to test the Solaris build on [current](https://downloads.omnios.org/media/stable/omnios-r151040g.iso) OmniOS (don't know how compatible they are really).

meson configures everything just fine, but code doesn't compile. Same with the 1.1 autotools branch. Some breakage seems to have accumulated since at least 7a54fe5e884e98ade91af527c67f9c5df1452a50.

I have zero experience with Solaris/Illumos. Maybe create an issue to restore support in case anyone knowledgeable is interested?

src/include/meson.build
```diff
@@ -0,0 +1,9 @@
+configure_file(output: 'config.h', configuration: cdata)
```
gsliepen, line 1: Just a question, why is this in it's own directory?

gsliepen (commented):
This looking very good! It's also very impressive to see it cross-build for Windows by just using `--cross-file .ci/cross/windows/amd64`, including all the wrapped dependencies it fetches. That will make it a lot easier for Windows users to build tinc themselves.

src/include/meson.build
hg: It's been a few days, I may misremember something, but I believe it goes like this.

With default settings, meson adds current source and build directories to include path.

They go right to the front of compiler's arguments: `cc -I. -I../src/foobar …`.

Under MinGW, `unistd.h` includes `process.h`. What it wants is to include the system header (`/mingw64/include/process.h`), but because of the argument order it actually includes `src/process.h` from tinc's sources, so we get an include cycle, and a compiler error.

Here's an [example test run](https://github.com/hg/tinc/runs/5611493644?check_suite_focus=true#step:7:142).

I [tried to fix it](https://github.com/hg/tinc/commit/068b1399cc78e25335ad5f4e59a622f772d02bc5) by renaming `src/process.h`, but thought that to be a pretty poor solution — who knows what conflicts we may get in the future, or on other operating systems?

So we have to disable this behavior for all targets with `implicit_include_directories: false`. From looking at other projects, I believe it's being used quite widely.

Generated headers go into the build directory. To be able to access them, you either have to use `implicit_include_directories: true` (which is the default), or `include_directories('.')`, which (if you put it into `src/meson.build`) pretty much overrides that setting and you get the `process.h` issue right back.

This is why it's moved into a separate subdirectory named `include`, which only contains the two generated headers and can be added to include path without adding other stuff (like `src/process.h`) that conflicts with system headers.

Thanks for the review. I'll fix everything else soon.

gsliepen:
> I tried to test the Solaris build on [current](https://downloads.omnios.org/media/stable/omnios-r151040g.iso) OmniOS (don't know how compatible they are really).
> 
> meson configures everything just fine, but code doesn't compile. Same with the 1.1 autotools branch. Some breakage seems to have accumulated since at least https://github.com/gsliepen/tinc/commit/7a54fe5e884e98ade91af527c67f9c5df1452a50.
> 
> I have zero experience with Solaris/Illumos. Maybe create an issue to restore support in case anyone knowledgeable is interested?

I just tried it with OpenIndiana, and it is so frustrating that it makes me want to throw my computer out of the window, even though it's just running in a VM. I think I will just remove Solaris support unless someone steps up to maintain it. So unless you really want to, don't spend any more effort on this.

.builds/openbsd.yml
```diff
@@ -1,34 +1,30 @@
...
-environment:
-  AUTOCONF_VERSION: 2.71
-  AUTOMAKE_VERSION: 1.16
-  CFLAGS: -I/usr/local/include -L/usr/local/lib
-
 sources:
   - https://github.com/gsliepen/tinc
 
+environment:
+  LD_LIBRARY_PATH: /usr/local/lib/eopenssl30
```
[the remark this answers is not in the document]
gsliepen: I just tried to build it on OpenBSD 7.0 myself, and I was surprised it didn't detect OpenSSL or LibreSSL support automatically, while this should just be part of the default install. It turns out it is there, and even has a pkg-config file, but OpenBSD's pkg-config file says it is openssl version 1.0.0.

It would be great if we do support OpenBSD with the default LibreSSL install, maybe a compromise is to detect if we are on OpenBSD, and then lower the minimum required version of OpenSSL to 1.0.0.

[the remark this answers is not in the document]
eli-schwartz: LibreSSL was forked from OpenSSL 1.0.x and its primary goal was to:
- remove support for all systems other than OpenBSD, since cross-platform support leads to code complexity
- remove legacy, deprecated ciphers rather than leaving them in as deprecated

Actually supporting OpenSSL 1.1.x APIs, which projects increasingly started using and depending on, was promised for years, while nothing ever really happened. And of course OpenBSD is not willing to accept Apache 2.0 code, so they cannot import changes from OpenSSL.

Anyway if you do not *require* (on an API level) openssl >=1.1.0 you don't need to version guard based on it. :)

src/meson.build
```diff
@@ -0,0 +1,404 @@
...
+  'netinet/in6.h',
+  'netinet/in_systm.h',
+  'netinet/ip.h',
+  'netinet/ip6.h',
+  'netinet/ip_icmp.h',
+  'netinet/tcp.h',
+  'netpacket/packet.h',
+  'resolv.h',
+  'stddef.h',
+  'sys/epoll.h',
```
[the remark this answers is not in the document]
hg: I moved a couple of other Linux-specific headers. All others can be found on opengroup.org and in various BSDs.

.builds/openbsd.yml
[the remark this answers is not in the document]
hg: LibreSSL seems to have support for all the APIs the project needs (unlike 'real' OpenSSL 1.0).

---

Empty fallback [doesn't work][ci] with meson 0.55, so I had to write this gem.

Hopefully we can start using the meta-dependency in a couple of years and the pain goes away.

[ci]: https://github.com/hg/tinc/runs/5650645312?check_suite_focus=true#step:7:173

hg:
[Full diff](https://github.com/gsliepen/tinc/compare/da0cc5a75ebbc294f8260f21f313ace7b548121f..df716df33af8e9a5b93d573a023ecd7fc24d9a03).

PR #356, opened 2022-03-23
Restore Solaris support

hg:
`scripts.test` is disabled on Solaris because it reliably produces somewhat different line ordering:

```
$ diff scripts.test.1/scripts.out scripts.test.1/scripts.out.expected
12d11
< bar-started-1
13a13
> bar-started-1
16d15
< bar-stopped
18a18
> bar-stopped
21d20
< bar-started-2
23a23
> bar-started-2
```

It's probably not worth it to spend any time on this. It was fun enough making it work on NetBSD — you adjust a couple of things, tests start passing there and failing everywhere else.

I also don't think it's possible to make tests work on native shell tools. For example, `grep` misses support for an important flag (`-m`) and doesn't provide any alternatives.

Thankfully the [system I'm using][omnios] has GNU coreutils preinstalled.

[omnios]: https://omnios.org/download.html

```
$ uname -a
SunOS omnios 5.11 omnios-r151040-d75907718a i86pc i386 i86pc
```
---

Did a bit of debugging. Looks like `tail` sometimes loses its position within a file when you append to it, and starts from the beginning. This makes `wait_script` completely useless. I haven't seen this behavior anywhere else.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -34,6 +34,10 @@ os_name = host_machine.system()
 cc_name = cc.get_id()
 
 cc_defs = ['-D_GNU_SOURCE']
+if os_name == 'sunos'
+  cc_defs += '-D__EXTENSIONS__'
+endif
+
 cc_flags = [cc_defs]
 ld_flags = []
 
--- a/src/fd_device.c
+++ b/src/fd_device.c
@@ -22,7 +22,6 @@
 
 #include "system.h"
 
-#ifdef HAVE_SYS_UN_H
 #include <sys/un.h>
 
 #include "conf.h"
@@ -240,4 +239,3 @@ const devops_t fd_devops = {
 	.read = read_packet,
 	.write = write_packet,
 };
-#endif
--- a/src/meson.build
+++ b/src/meson.build
@@ -46,23 +46,31 @@ check_headers = [
   'sys/stat.h',
   'sys/time.h',
   'sys/types.h',
-  'sys/un.h',
   'sys/wait.h',
   'syslog.h',
   'termios.h',
 ]
 
+# 'struct msghdr' misses some required fields
+if os_name != 'sunos'
+  check_headers += 'sys/un.h'
+endif
+
 check_functions = [
   'asprintf',
   'daemon',
   'fchmod',
-  'fork',
   'gettimeofday',
   'mlockall',
   'putenv',
   'strsignal',
   'unsetenv',
 ]
+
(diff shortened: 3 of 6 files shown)


test/testlib.sh
```diff
@@ -45,6 +45,21 @@ DIR_BAZ=$(realdir "$PWD/$net3")
...
+  tail() { $gnu/tail "$@"; }
+
+  if ! tail /dev/null || ! echo '' | grep ''; then
+    echo >&2 'Sorry, native Solaris tools are not supported. Please install GNU Coreutils.'
+    exit $EXIT_SKIP_TEST
+  fi
+else
+  grep='grep'
+fi
+
```
gsliepen, line 62: As @eli-schwartz mentioned in #357, meson already depends on Python, so that makes me wonder if we should not just rewrite all the test scripts in Python. But that's something for another pull request.

eli-schwartz: In theory there is also https://sr.ht/~lattis/muon /https://muon.build which is a c99 rewrite of Meson with a minimal bootstrap path and suitable for addressing https://mesonbuild.com/FAQ.html#but-i-really-want-a-version-of-meson-that-doesnt-use-python

Not everyone cares about that, of course. Practically speaking, depending on python for the build machine isn't so unreasonable. And muon doesn't care if you use scripts which depend on python, it will still work (at least if you don't use muon's current major weak point which is modules support). But you will of course no longer be able to depend on muon allowing you to knock python out of the build requirements.

I'd consider that benefit to be mainly of interest to people doing systems bootstrapping especially in cases where you are operating a self-hosting Linux distro where base features like the init system or C library use Meson to build. That's actually a major consideration for Alpine Linux and OpenRC.

I don't know if you care. Probably not, as tinc isn't really a distro bootstrap package.

hg: The idea already came up when we were improving tests last year, but was rejected because Python wasn't required back then.

Unix utilities are very similar… from a 10 000-foot view. When you start doing anything complicated, you run into lots of smalls differences.

So tests already have "non-native" dependencies because I couldn't make them work reliably everywhere: procps, coreutils, a particular flavor of netcat.

I think I'd rather replace all that with Python.

eli-schwartz: Yeah, making things work for the least common denominator of "POSIX specifies this utility with this behavior" can be quite hard to build complex actions out of.

The only other alternative is probably to write the tests in C.

PR #357, opened 2022-03-25
Add support for building with MSVC

hg:
Integration tests are disabled because of their dependence on Unix-like environment.

~~Project version will either have to be updated manually from now on, or you could start using tags without the `release-` prefix (then let me know and I'll update `meson.build`). We need`'release-1.18'.replace('release-', '')` to work, which only became available [in 0.58](https://mesonbuild.com/Syntax.html#replace), and you can't check what OS you're building for before `project()` is finished, so there's no way to ship an alternative script for Windows.~~

All compiler attributes are ignored for now, except for `packed`. I just realized we're checking for attribute support, but not doing anything with that information (same thing was happening with autotools). I think it's best to fix this in a separate PR.

---

mac had a random failure. NetBSD looks suspicious. I'll test it more locally.

### DIFF
--- a/.ci/windows/build.cmd
+++ b/.ci/windows/build.cmd
@@ -1,0 +1,8 @@
+set crypto=%1
+set builddir=%crypto%
+
+echo configure build directory
+meson setup %builddir% -Dbuildtype=release -Dcrypto=%crypto% || exit 1
+
+echo build project
+meson compile -C %builddir% || exit 1
--- a/meson.build
+++ b/meson.build
@@ -1,5 +1,5 @@
 project('tinc', 'c',
-  version: '1.18pre',
+  version: run_command([find_program('python3'), 'version.py', 'short'], check: true).stdout(),
   license: 'GPL-2.0-or-later',
   meson_version: '>=0.51',
   default_options: [
@@ -34,6 +34,13 @@ cc = meson.get_compiler('c')
 os_name = host_machine.system()
 cpu_family = host_machine.cpu_family()
 cc_name = cc.get_id()
+
+python = find_program('python3')
+if meson_version.version_compare('>=0.55')
+  python_path = python.full_path()
+else
+  python_path = python.path()
+endif
 
 cc_defs = ['-D_GNU_SOURCE']
 if os_name == 'sunos'
--- a/meson.build
+++ b/meson.build
@@ -1,5 +1,5 @@
 project('tinc', 'c',
-  version: run_command('./src/git_tag.sh', check: true).stdout().strip(),
+  version: '1.18pre',
   license: 'GPL-2.0-or-later',
   meson_version: '>=0.51',
   default_options: [
@@ -12,6 +12,7 @@ project('tinc', 'c',
 dir_run_state = get_option('runstatedir')
 opt_crypto = get_option('crypto')
 opt_curses = get_option('curses')
+opt_debug = get_option('debug')
 opt_docs = get_option('docs')
 opt_harden = get_option('hardening')
 opt_jumbograms = get_option('jumbograms')
@@ -31,6 +32,7 @@ meson_version = meson.version()
 
 cc = meson.get_compiler('c')
 os_name = host_machine.system()
+cpu_family = host_machine.cpu_family()
 cc_name = cc.get_id()
 
 cc_defs = ['-D_GNU_SOURCE']
@@ -47,33 +49,45 @@ else
   static = opt_static.enabled()
 endif
 
-if static
+if static and cc_name != 'msvc'
   ld_flags += '-static'
 endif
 
 if opt_harden
-  cc_flags += [
-    '-D_FORTIFY_SOURCE=2',
-    '-fwrapv',
-    '-fno-strict-overflow',
-    '-Wreturn-type',
-    '-Wold-style-definition',
-    '-Wmissing-declarations',
-    '-Wmissing-prototypes',
-    '-Wstrict-prototypes',
-    '-Wredundant-decls',
-    '-Wbad-function-cast',
-    '-Wwrite-strings',
-    '-fdiagnostics-show-option',
-    '-fstrict-aliasing',
-    '-Wmissing-noreturn',
-  ]
-  if cc_name == 'clang'
-    cc_flags += '-Qunused-arguments'
-  endif
-  ld_flags += ['-Wl,-z,relro', '-Wl,-z,now']
-  if os_name == 'windows'
-    ld_flags += ['-Wl,--dynamicbase', '-Wl,--nxcompat']
+  if cc_name == 'msvc'
+    # Most of these flags are already ON by default in the latest version of MSVC.
+    # Add anyway in case someone is building using an old toolchain.
+    cc_flags += ['/guard:cf', '/GS']
+    ld_flags += [
+      '/guard:cf',
+      '/NXCOMPAT',
+      '/DYNAMICBASE',
+      cpu_family.endswith('64') ? '/HIGHENTROPYVA' : '/SAFESEH',
+    ]
+  else
+    cc_flags += [
+      '-D_FORTIFY_SOURCE=2',
+      '-fwrapv',
+      '-fno-strict-overflow',
+      '-Wreturn-type',
+      '-Wold-style-definition',
+      '-Wmissing-declarations',
+      '-Wmissing-prototypes',
+      '-Wstrict-prototypes',
+      '-Wredundant-decls',
+      '-Wbad-function-cast',
+      '-Wwrite-strings',
+      '-fdiagnostics-show-option',
+      '-fstrict-aliasing',
+      '-Wmissing-noreturn',
+    ]
+    if cc_name == 'clang'
+      cc_flags += '-Qunused-arguments'
+    endif
+    ld_flags += ['-Wl,-z,relro', '-Wl,-z,now']
+    if os_name == 'windows'
+      ld_flags += ['-Wl,--dynamicbase', '-Wl,--nxcompat']
+    endif
   endif
 endif
 
--- a/src/ethernet.h
+++ b/src/ethernet.h
@@ -68,6 +68,8 @@ struct ether_header {
 };
 #endif
 
+STATIC_ASSERT(sizeof(struct ether_header) == 14, "ether_header has incorrect size");
+
 #ifndef HAVE_STRUCT_ARPHDR
 struct arphdr {
 	uint16_t ar_hrd;
@@ -86,6 +88,8 @@ struct arphdr {
 #define ARPOP_NAK 10
 #endif
 
+STATIC_ASSERT(sizeof(struct arphdr) == 8, "arphdr has incorrect size");
+
 #ifndef HAVE_STRUCT_ETHER_ARP
 struct  ether_arp {
 	struct  arphdr ea_hdr;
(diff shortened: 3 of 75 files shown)


eli-schwartz:
> and you can't check what OS you're building for before `project()` is finished, so there's no way to ship an alternative script for Windows.

You don't need a batch file on Windows and a shell script on unix.

You could use a python script on both, relying on the fact that Meson itself is written in python and can use that as a fallback to run the script.

hg:
Turns out you can just call the script and meson will wire everything up, even if there's no `python` in your `%PATH%`. I was expecting the opposite for some reason. Thank you, @eli-schwartz.

src/windows/device.c
```diff
@@ -1,5 +1,5 @@
 /*
-    device.c -- Interaction with Windows tap driver in a MinGW environment
+    device.c -- Interaction with Windows tap driver
```
gsliepen, line 2: Maybe write "Interaction with the TAP-Windows driver"

src/invitation.c
```diff
@@ -662,7 +664,7 @@ static bool finalize_join(void) {
 	}
 
 	size_t len = strlen(temp_name);
-	char name[len + 1];
+	char *name = alloca(len + 1);
```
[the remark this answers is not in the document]
hg: `get_value()` calls `get_line()`, which uses its own static buffer for storing text.

It also gets called later in a loop. "Join" tests are not happy:

```
Ignoring unknown variable '-----END' in invitation.
Ignoring unsafe variable 'Ed25519PublicKey' in invitation.
Ignoring unsafe variable 'Port' in invitation.
Ignoring unsafe variable 'Address' in invitation.
```

`get_line()` now takes buffer as an argument to make this a bit more obvious.

src/ethernet.h
```diff
@@ -61,21 +61,21 @@
 #endif
 
 #ifndef HAVE_STRUCT_ETHER_HEADER
-struct ether_header {
+PACKED(struct ether_header {
```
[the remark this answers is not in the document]
hg: Checked what others are doing just to be sure.

[FreeBSD][fbsd] and [Linux][linux] are using `__packed__`, while [OpenBSD][obsd], [NetBSD][nbsd] and [Illumos][sun] are not.

So we're probably fine without it.

[fbsd]: https://github.com/freebsd/freebsd-src/blob/e51be99e8ad021923e74c9a9e1b87c2c2309b561/sys/net/ethernet.h#L62-L66
[linux]: https://elixir.bootlin.com/linux/v5.17/source/include/uapi/linux/if_ether.h#L169
[obsd]: https://github.com/openbsd/src/blob/baf715768bffa92c878b78653d9c82e7e7dc7978/sys/netinet/if_ether.h#L77-L81
[nbsd]: https://github.com/NetBSD/src/blob/trunk/sys/net/if_ether.h#L87-L91
[sun]: https://github.com/illumos/illumos-gate/blob/6f443ebc1fb4fec01d6e8fa8ca4648182ed215bb/usr/src/uts/common/sys/ethernet.h#L47-L63

src/invitation.c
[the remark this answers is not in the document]
gsliepen: Oops, you're right.

src/ethernet.h
```diff
@@ -65,17 +65,17 @@ struct ether_header {
 	uint8_t ether_dhost[ETH_ALEN];
 	uint8_t ether_shost[ETH_ALEN];
 	uint16_t ether_type;
-} __attribute__((__gcc_struct__)) __attribute((__packed__));
+};
 #endif
 
 #ifndef HAVE_STRUCT_ARPHDR
-struct arphdr {
+PACKED(struct arphdr {
```
gsliepen, line 72: Still some `PACKED`s? I think they can *all* be removed.

hg: Ah, sorry, you're right. [OpenBSD][one] still [uses][two] them [for some][thr] structs, but it doesn't seem necessary:

[one]: https://github.com/openbsd/src/blob/2207c4325726fdc5c4bcd0011af0fdf7d3dab137/sys/netinet/ip6.h#L72
[two]: https://github.com/openbsd/src/blob/2207c4325726fdc5c4bcd0011af0fdf7d3dab137/sys/netinet/icmp6.h#L70
[thr]: https://github.com/openbsd/src/blob/2207c4325726fdc5c4bcd0011af0fdf7d3dab137/sys/netinet/icmp6.h#L212

https://github.com/NetBSD/src/commit/427831ba4bdf388aecf3a378de8faf3a4d44a462


I rearranged lines into their original position, the diff is smaller than it looks:

    $ git diff upstream/1.1..HEAD src/{ipv4,ipv6,ethernet}.h

```diff
diff --git a/src/ethernet.h b/src/ethernet.h
index 3b074b7f..4ef9dab9 100644
--- a/src/ethernet.h
+++ b/src/ethernet.h
@@ -65,7 +65,7 @@ struct ether_header {
 	uint8_t ether_dhost[ETH_ALEN];
 	uint8_t ether_shost[ETH_ALEN];
 	uint16_t ether_type;
-} __attribute__((__gcc_struct__)) __attribute((__packed__));
+};
 #endif

 #ifndef HAVE_STRUCT_ARPHDR
@@ -75,7 +75,7 @@ struct arphdr {
 	uint8_t ar_hln;
 	uint8_t ar_pln;
 	uint16_t ar_op;
-} __attribute__((__gcc_struct__)) __attribute((__packed__));
+};

 #define ARPOP_REQUEST 1
 #define ARPOP_REPLY 2
@@ -93,7 +93,7 @@ struct  ether_arp {
 	uint8_t arp_spa[4];
 	uint8_t arp_tha[ETH_ALEN];
 	uint8_t arp_tpa[4];
-} __attribute__((__gcc_struct__)) __attribute((__packed__));
+};
 #define arp_hrd ea_hdr.ar_hrd
 #define arp_pro ea_hdr.ar_pro
 #define arp_hln ea_hdr.ar_hln
diff --git a/src/ipv4.h b/src/ipv4.h
index 708e5c1f..a412c605 100644
--- a/src/ipv4.h
+++ b/src/ipv4.h
@@ -83,7 +83,7 @@ struct ip {
 	uint8_t ip_p;
 	uint16_t ip_sum;
 	struct in_addr ip_src, ip_dst;
-} __attribute__((__gcc_struct__)) __attribute((__packed__));
+};
 #endif

 #ifndef IP_OFFMASK
@@ -145,7 +145,7 @@ struct icmp {
 #define icmp_radv icmp_dun.id_radv
 #define icmp_mask icmp_dun.id_mask
 #define icmp_data icmp_dun.id_data
-} __attribute__((__gcc_struct__)) __attribute((__packed__));
+};
 #endif

 #endif
diff --git a/src/ipv6.h b/src/ipv6.h
index cfa6f239..7c06de3d 100644
--- a/src/ipv6.h
+++ b/src/ipv6.h
@@ -51,7 +51,7 @@ struct ip6_hdr {
 	} ip6_ctlun;
 	struct in6_addr ip6_src;
 	struct in6_addr ip6_dst;
-} __attribute__((__gcc_struct__)) __attribute((__packed__));
+};
 #define ip6_vfc ip6_ctlun.ip6_un2_vfc
 #define ip6_flow ip6_ctlun.ip6_un1.ip6_un1_flow
 #define ip6_plen ip6_ctlun.ip6_un1.ip6_un1_plen
@@ -70,7 +70,7 @@ struct icmp6_hdr {
 		uint16_t icmp6_un_data16[2];
 		uint8_t icmp6_un_data8[4];
 	} icmp6_dataun;
-} __attribute__((__gcc_struct__)) __attribute((__packed__));
+};
 #define ICMP6_DST_UNREACH_NOROUTE 0
 #define ICMP6_DST_UNREACH 1
 #define ICMP6_PACKET_TOO_BIG 2
@@ -90,7 +90,7 @@ struct icmp6_hdr {
 struct nd_neighbor_solicit {
 	struct icmp6_hdr nd_ns_hdr;
 	struct in6_addr nd_ns_target;
-} __attribute__((__gcc_struct__)) __attribute((__packed__));
+};
 #define ND_OPT_SOURCE_LINKADDR 1
 #define ND_OPT_TARGET_LINKADDR 2
 #define nd_ns_type nd_ns_hdr.icmp6_type
@@ -103,7 +103,7 @@ struct nd_neighbor_solicit {
 struct nd_opt_hdr {
 	uint8_t nd_opt_type;
 	uint8_t nd_opt_len;
-} __attribute__((__gcc_struct__)) __attribute((__packed__));
+};
 #endif

 #endif
```

hg:
After further testing on mac, [this](https://github.com/gsliepen/tinc/runs/5711402692?check_suite_focus=true) seems to be a random failure. They happen much more frequently on mac than on other operating systems. Since tests are being rewritten anyway, we'll see if it continues (maybe it's a bug in tinc then).

gsliepen:
I want to see if I can rewrite `security.test` and `legacy-protocol.test` anyway; they take much longer to run than the other tests. Currently they rely a lot on timing.

hg:
We probably can't get rid of `packed` because of Windows, unless I'm doing something wrong again: [msvc][msvc], [gcc][gcc].

The [change][commit] in question.

[msvc]: https://github.com/hg/tinc/runs/5715654457?check_suite_focus=true#step:5:648
[gcc]: https://github.com/hg/tinc/runs/5715654085?check_suite_focus=true#step:7:258
[commit]: https://github.com/hg/tinc/commit/5f5ad04a8fae2a2e65682877cafbf42c167d0395

I'll look into it more a bit later.

.ci/windows/build.cmd
```diff
@@ -0,0 +1,13 @@
+set crypto=%1
```
gsliepen, line 1: It's a bit weird to call the build directory "crypto".

.ci/windows/build.cmd
```diff
@@ -0,0 +1,13 @@
...
+meson setup %crypto% -Dbuildtype=release -Dcrypto=%crypto% || exit 1
+meson compile -C %crypto% || exit 1
+
+mkdir %data% || exit 1
+
+%crypto%\src\tinc --version || exit 1
+%crypto%\src\tinc -c %data% -b init foo || exit 1
+%crypto%\src\tinc -c %data% -b generate-ed25519-keys || exit 1
+
+%crypto%\src\tincd --version || exit 1
```
gsliepen, line 13: This looks like a test, I don't think it should be part of the build process.

gsliepen (commented):
Lots of unavoidable churn due to the lack of `__attribute__`, `s/MINGW/WINDOWS`, but apart from that it looks quite good.

meson.build
```diff
@@ -31,8 +32,16 @@ meson_version = meson.version()
...
 os_name = host_machine.system()
+cpu_family = host_machine.cpu_family()
 cc_name = cc.get_id()
 
+python = find_program('python3')
+if meson_version.version_compare('>=0.55')
+  python_path = python.full_path()
+else
+  python_path = python.path()
+endif
```
eli-schwartz, line 43: I neglected to re-review this before, but... why do you need this?

src/include/meson.build
```diff
@@ -1,7 +1,7 @@
 configure_file(output: 'config.h', configuration: cdata)
 
 src_lib_common += vcs_tag(
-  command: './git_tag.sh',
+  command: [python_path, src_root / 'version.py'],
```
eli-schwartz, line 4: You use the path to python here, but command lists are supposed to accept found programs here almost per definition, because every command begins with a program (and continues with optional args).

hg: Well, it doesn't here… tried that, of course.

meson.build
hg: For `vcs_tag`, and also knowing that I'll be needed for tests.

PR #367, opened 2022-04-11
Use struct for legacy proto fields in connection_t

hg:
I remember reading quite a while ago that legacy protocol stuff in `connection_t` should be separated into another struct. Was this what you meant?

### DIFF
--- a/src/connection.h
+++ b/src/connection.h
@@ -27,6 +27,7 @@
 #include "rsa.h"
 #include "list.h"
 #include "sptps.h"
+#include "logger.h"
 
 #define OPTION_INDIRECT         0x0001
 #define OPTION_TCPONLY          0x0002
@@ -59,6 +60,7 @@ typedef union connection_status_t {
 #include "edge.h"
 #include "net.h"
 #include "node.h"
+#include "compression.h"
 
 #ifndef DISABLE_LEGACY
 typedef struct legacy_crypto_t {
@@ -107,7 +109,11 @@ typedef struct connection_t {
 	sptps_t sptps;
 
 	int outmaclength;
-	int outcompression;             /* compression level from compression_level_t */
+
+	union {
+		compression_level_t outcompression;
+		debug_t log_level; // used for REQ_LOG
+	};
 
 	uint8_t *hischallenge;          /* The challenge we sent to him */
 	uint8_t *mychallenge;           /* The challenge we received */
--- a/src/connection.h
+++ b/src/connection.h
@@ -67,7 +67,7 @@ typedef struct legacy_crypto_t {
 	uint64_t budget;
 } legacy_crypto_t;
 
-bool init_crypto_by_nid(legacy_crypto_t *c, int cipher, int digest) ATTR_WARN_UNUSED;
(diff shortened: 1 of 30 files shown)


src/connection.h
```diff
@@ -79,13 +93,7 @@ typedef struct connection_t {
...
 
 #ifndef DISABLE_LEGACY
-	rsa_t *rsa;                     /* his public RSA key */
-	cipher_t incipher;              /* Cipher he will use to send data to us */
-	cipher_t outcipher;             /* Cipher we will use to send data to him */
-	digest_t indigest;
-	digest_t outdigest;
-	uint64_t inbudget;
-	uint64_t outbudget;
+	legacy_ctx_t legacy;
```
gsliepen, line 96: Yes, but the idea was also to make this a pointer, so we can opt not to allocate memory for it if we are not using the legacy protocol to communicate with the peer.

hg: Right, this makes more sense.

On an unrelated note, should I send in changes like [this one]?

[types]: https://github.com/hg/tinc/commit/5697f403b900989e68d9b0309ff0a54341ea9dda

Strictly speaking, it's quite useless since C type system is so primitive, but it at least makes intent more clear.

gsliepen: Sure, I will merge patches like [that one](https://github.com/hg/tinc/commit/5697f403b900989e68d9b0309ff0a54341ea9dda). Although I wouldn't recommend spending too much time one things like this if you think there is more important stuff to do :)

hg: Sure, I don't spend time specifically on such stuff (there's not very much of it already). I replaced a couple of nids while working on this PR, thinking it will involve only one or two lines, but it turned out to be quite a bit larger than that.

Let's just throw that commit on top of this PR then, since it depends on it (unless it gets in the way).

src/protocol_key.c
```diff
@@ -402,14 +403,14 @@ bool ans_key_h(connection_t *c, const char *request) {
...
+	nid_t cipher, digest;
 	unsigned long maclength;
-	int compression;
+	compression_level_t compression;
 	node_t *from, *to;
 
 	if(sscanf(request, "%*d "MAX_STRING" "MAX_STRING" "MAX_STRING" %d %d %lu %d "MAX_STRING" "MAX_STRING,
 	                from_name, to_name, key, &cipher, &digest, &maclength,
-	                &compression, address, port) < 7) {
+	                (int *)&compression, address, port) < 7) {
```
gsliepen, line 413: I've been bitten by code like this in the past. It's safer to keep the local variables `int`, read into those, and then cast it to `nid_t`/`compression_level_t` later.

hg: Right, thank you for having a look at it.

I think `compression_level` in `connection_t` isn't actually used anymore, and its comparison with debug level here felt a bit weird:

https://github.com/gsliepen/tinc/blob/1.1/src/logger.c#L113-L115

so I split it into two fields and then removed the old one.

gsliepen:
Rebased as 391985158c16f4a3894a798f6788cfb23dfcfea9

PR #368, opened 2022-04-11
Changes to cryptographic RNG

hg:
- extract copypasted code into a separate file
- use it for libgcrypt (since we're already using it for OpenSSL, even though it provides its own generator)
- add support for `getrandom()` on Linux

`getrandom` doesn't add very much, but it's also practically free to support; takes one open file descriptor away, adds the ability to work in badly configured chroots/containers where `/dev/(u)random` is missing.


test/unit/test_random.c
```diff
@@ -0,0 +1,54 @@
...
+	randomize(data, sizeof(data));
+
+	assert_memory_not_equal(ref, data, sizeof(data));
+}
+
+int main(void) {
+	const struct CMUnitTest tests[] = {
+		cmocka_unit_test(test_randomize_zero_must_not_change_memory),
+		cmocka_unit_test(test_randomize_half_changes_half),
+		cmocka_unit_test(test_randomize_full_changes_memory),
```
gsliepen, line 51: I think it might then also be nice that `randomize()` produces different data every time it is called.

src/gcrypt/crypto.c
```diff
@@ -17,18 +17,13 @@
...
 
-#include "../system.h"
-
-#include <gcrypt.h>
-
+#include "../random.h"
 #include "../crypto.h"
 
 void crypto_init(void) {
+	random_init();
```
gsliepen, line 24: I think the crypto functions shouldn't call `random_init()` and `random_exit()` anymore then, as that's still a bit of duplication.

test/unit/test_random.c
hg: I thought about adding statistical tests, but it's a bit too much. This is good middle ground.

Also added a check that `randomize()` fails if you forget to initialize it.

The best way to do that would probably be [mocking `assert`][assert], but it requires rebuilding the project with a different set of compiler flags.

[assert]: https://api.cmocka.org/group__cmocka__mock__assert.html

Current meson configs don't allow that, so it's a relatively big task. I'll look into that separately.

I tried a bunch of other hacks (handling `SIGABRT`, mocking `abort()`, mocking a wrapper around `abort()`, `longjmp` trickery, `fork` + `wait`). Those turned out to be ugly for one reason or another.

If you'd rather wait for a proper solution, I'll remove the test for now.

src/gcrypt/crypto.c
hg: Calls to `crypto_exit()` were missing in some places, so I had to rearrange some code to avoid repeating `random_exit` before two dozen `return`s.

PR #369, opened 2022-04-13
Convert tincd path args to absolute paths

hg:
Ran into this many times when testing tinc manually:

    $ mkdir x
    $ tinc -c x init foo
    $ tinc -c x set DeviceType dummy
    $ tincd -c x -D
    Cannot open config file x/tinc.conf: No such file or directory
    Failed to read `x/tinc.conf': No such file or directory

while absolute paths work fine:

    $ tincd -c $PWD/x -D

It happens because `tincd` stores paths as they were passed, then does `chdir` to its configuration directory, and then tries to `fopen`.

I went for the simplest solution — resolve all paths to absolute before doing `chdir`.

Sadly, `realpath` fails if path does not exist (which is a problem for pidfile and logfile), so if that happens we fall back to concatenating `getcwd` and whatever was passed by the user.

It's what [others][cwalk] are using, with the exception that we're skipping normalization (`fopen` works either way, and I don't think it makes sense to drag so much complexity to make logs a bit prettier).

[cwalk]: https://github.com/likle/cwalk/blob/3aa1e36aa427c33351e7120b56e95f69e432d683/src/cwalk.c#L647

Not sure about logfile and pidfile. If you pass relative paths, they currently work as relative to configuration directory, although it isn't obvious and is not mentioned in the documentation. With this PR it's a bit more consistent, but may cause regressions (although 1.1 is officially a development release).

What are your thoughts? We could use the old behavior for these two, and get rid of `absolute_path` (`realpath` will do since configuration directory must exist before starting `tincd`).

### DIFF
--- a/src/tincd.c
+++ b/src/tincd.c
@@ -158,6 +158,17 @@ static void usage(bool status) {
 	}
 }
 
+// Try to resolve path to absolute, return a copy of the argument if this fails.
+static char *get_path_arg(char *arg) {
+	char *result = absolute_path(arg);
+
+	if(!result) {
+		result = xstrdup(arg);
+	}
+
+	return result;
+}
+
 static bool parse_options(int argc, char **argv) {
 	config_t *cfg;
 	int r;
@@ -175,7 +186,7 @@ static bool parse_options(int argc, char **argv) {
 
 		case OPT_CONFIG_FILE:
 			free(confbase);
-			confbase = xstrdup(optarg);
+			confbase = get_path_arg(optarg);
 			break;
 
 		case OPT_NO_DETACH:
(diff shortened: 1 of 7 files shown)


gsliepen:
Your solution fails if you also use the `--chroot` option (not that it worked well before).

There are two reasons why we `chdir()`:
1. Ensure we don't keep an arbitrary mount point busy
2. Ensure scripts are started with a predictable working directory

I think the best way would be to open the pidfile, logfile and control socket before calling `chdir()`, then after `chdir()` open `tinc.conf` and other config files without prepending `confbase` to it, because we know we already `chdir()`'ed into the configuration directory. The only question is how this will interact with `daemon()`, or worse, Windows's way of starting a service.

gsliepen:
Hm, this is more difficult than I thought. Opening files before `chdir()`/`chroot()` is not enough; we want to be able to clean up PID files if we quit, and reopen log files if there's a HUP signal (to allow log rotation). We also have the daemon doing `chdir()` but the CLI not, and some code that builds path names is shared between them. So making absolute paths looks like the best solution after all.

PR #370, opened 2022-04-16
Wipe (some) secrets from memory after use

hg:
While not a panacea, this is considered to be a good practice which doesn't cost us much.

The cryptographic libraries we're using are already doing that for some of their own data structures. This should additionally cover EC keys, a few strings potentially containing passwords, and various temporary key buffers.

---

I finally remembered to add libgcrypt to CI. This is not related to the main PR, but is bundled here to avoid creating even more conflicts.

### DIFF
--- a/.ci/sanitizers/run.sh
+++ b/.ci/sanitizers/run.sh
@@ -8,13 +8,14 @@ logs="$GITHUB_WORKSPACE/sanitizer"
 
 case "$SANITIZER" in
 undefined)
-  flags='-fsanitize=integer -fsanitize=nullability'
+  flags='-fsanitize=integer -fsanitize=nullability -fno-sanitize=unsigned-integer-overflow'
   export UBSAN_OPTIONS="log_path=$logs/ubsan:print_stacktrace=1"
   ;;
 
 address)
   flags='-fsanitize-address-use-after-scope -fsanitize=pointer-compare -fsanitize=pointer-subtract'
   export ASAN_OPTIONS="log_path=$logs/asan:detect_invalid_pointer_pairs=2:strict_string_checks=1:detect_stack_use_after_return=1:check_initialization_order=1:strict_init_order=1"
+  export LSAN_OPTIONS="suppressions=$dir/suppress.txt:print_suppressions=0"
   ;;
 
 thread)
--- a/.ci/sanitizers/suppress.txt
+++ b/.ci/sanitizers/suppress.txt
@@ -1,0 +1,4 @@
+# known leak from one-time allocation
+# upstream refuses to fix such things: https://dev.gnupg.org/T4499
+leak:libgcrypt.so
+
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -120,12 +120,18 @@ jobs:
 
       - name: Run tests with OpenSSL 3
         run: bash .ci/sanitizers/run.sh openssl3
+        if: always()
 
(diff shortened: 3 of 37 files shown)


src/splay_tree.c
```diff
@@ -275,7 +275,9 @@ splay_tree_t *splay_alloc_tree(splay_compare_t compare, splay_action_t delete) {
 }
 
 void splay_free_tree(splay_tree_t *tree) {
-	free(tree);
+	if(tree) {
```
gsliepen, line 278: This is unnecessary, you can call `free()` on a NULL-pointer without problems.

gsliepen (commented):
Looks OK. Although perhaps you might want to add a counterpart to `xzalloc()`:
```
void xzfree(void *buf, size_t len) {
    if(buf) {
        memzero(buf, len);
        free(buf);
    }
}
```
Since that pattern is now used a lot.

hg:
Good idea, thanks. I also added [`explicit_memset()`][ms] since NetBSD decided to be a special snowflake in that regard. Now the fallback shouldn't be used by anything.


[ms]: https://man.netbsd.org/NetBSD-8.0/explicit_memset.3


---



Another drive-by question: I've been updating man pages recently, and not having had the chance to write roff in the past, it's been… a bit of a rough ride. Macro names are really non-obvious, whitespace rules are somewhat arcane, and there are basically no modern tutorials for those who don't like to read (colorful, glossy, and and ridden with emoji, like you'd expect in the 21st century).

Now, I don't care about that much (most of the pain has been overcome by now), but if we hope to attract more contributors in the coming years, maybe it makes sense to look for more modern alternatives?

Like [scdoc], which provides markdown-ish syntax, compiles into roff, is written in C (and is tiny), and available everywhere.

[scdoc]: https://git.sr.ht/~sircmpwn/scdoc/

I mean, compare [this][roff] to [this][sc1] or [this][sc2] (or even [this][md], since it resembles markdown enough that GitHub markdown renderer mistakes it for one).

[md]: https://github.com/swaywm/sway/blob/master/sway/sway.1.scd
[roff]: https://raw.githubusercontent.com/gsliepen/tinc/1.1/doc/tinc.conf.5.in
[sc1]: https://raw.githubusercontent.com/swaywm/sway/master/sway/sway.5.scd
[sc2]: https://raw.githubusercontent.com/swaywm/sway/master/sway/sway.1.scd

Now, markdown can be converted into man pages, but that requires something like pandoc, which is pretty heavy, while scdoc binary package on my system is around 30 KBs.

The downside is obvious — it adds another build dependency (although we can write scdoc, convert it to roff, and add that roff to the source tree).

Yes/no?

gsliepen:
Looks good!

---

Tinc uses the [mdoc](https://www.freebsd.org/cgi/man.cgi?mdoc(7)) macros, which is already a higher level than pure roff, and intends to convey semantics instead of just how you want to style things (think LaTeX vs. TeX). In the past I was not too happy with the existing manpage generators, but perhaps they are better today.

As for scdoc, it looks cute, but whether it is tiny or not doesn't make much of a difference to someone who needs to install it as yet another dependency. I personally also find it very irritating that it's *almost* Markdown but didn't just go all the way. OTOH, I see scdoc is available in many package managers, which is great. If we go this route then maybe we can consider writing the manual in Markdown or something similar as well. Even more ideal would be to avoid the duplication of the documentation of command line and config file options.

Issue #363, opened 2022-04-08
Report actual listening port when using `Port = 0`


gsliepen:
If we let the OS choose a listening port, it would be nice to ensure the actual port number ends up being printed when tincd starts, is sent in meta-protocol `ACK` messages, is used in invitation URLs and invitation files, and can be queried using `tinc`.

gsliepen:
Fixed in 0289162552cd85375605044c696e2a3294e7aa9a

PR #371, opened 2022-04-17
Handle `Port 0`: use actual port in tincd logs / tinc get Port / invitation URLs and files

hg:
This should cover #363.

Because TCP and UDP ports can differ if `Port 0` is used, `myport` was converted to a struct.

Since communication with tincd involves parsing the pidfile, we read the port from there instead of asking tincd. This is simpler, and does not requre adding more control request types, breaking compatibility with older daemons.

I had to write a small piece of yet another configuration parser. I'll look into extracting them into a separate 'library' so we don't reimplement the wheel over and over again.

### DIFF
--- a/src/control.c
+++ b/src/control.c
@@ -30,6 +30,7 @@
 #include "utils.h"
 #include "xalloc.h"
 #include "random.h"
+#include "pidfile.h"
 
 char controlcookie[65];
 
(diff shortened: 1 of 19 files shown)


gsliepen:
This is already an improvement. However, if I start tinc with `Port 0` it currently is listening on 4 different ports, because of having both TCP and UDP sockets on both IPv4 and IPv6. Part of the protocol assumes that there is just one port tinc is listening in for everything. With two daemons listening on `Port 0`, with your patch it would be possible to have one invite the other via IPv4. However, if they were both dual-stacked but there was no viable IPv4 connection between them, only IPv6, they would not be able to join. Also, UDP would fail to work.

A possible solution is to check after binding the first TCP listening port what the actual port is (this can be done using `getsockname()`). Then it can use that port for all other sockets. If a subsequent listening socket fails to bind, you could still fall back to port 0 for that socket.

hg:
Good idea. This should cover it.

    udp   UNCONN 0      0      0.0.0.0:35887      0.0.0.0:*
    udp   UNCONN 0      0         [::]:35887         [::]:*
    tcp   LISTEN 0      3      0.0.0.0:35887      0.0.0.0:*
    tcp   LISTEN 0      3         [::]:35887         [::]:*




re CI failure: ugh. I wish there was an active macOS user among the developers. Apple really doesn't want you to support their OS.

GitHub CI has a 6-hour time limit, and you can SSH in there through third-party hacks like [this one][ssh]. I'll see that'd be enough to debug the issues we've been having.

[ssh]: https://github.com/mxschmitt/action-tmate

gsliepen:
I have a Mac Mini I tested it on.

PR #372, opened 2022-04-18
Changes to compiler flags & function attributes

hg:
Bits and pieces that are somewhat related:

- change some compiler flags:
    - enable CFI checks (no effect on older CPUs, does [this][cfi] on modern x86; the article is about the kernel, but the tech is similar). Ubuntu has been doing that [for years][cet].
    - add more stack smashing checks (and make stack non-executable)
    - enable more warnings on top of `-Wall -Wextra` (particularly warn about VLAs which we cannot use anymore because of MSVC, and about missing enum `case`s, which helps when adding new values to enums)
- fix warnings triggered by new flags
- since doing that involves adding a bunch of `__attribute__((format))`, use the opportunity to add `__attribute__((malloc(dealloc))` so [newer GCC][dealloc] warn us about using wrong deallocators
- fix a couple of leaks found by `-fanalyzer` (gives lots of false positives, so it's a bit early to enable it permanently)
- use current year in copyright messages

[dealloc]: https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html

Currently `tinc --version` says the copyright expired 3 years ago, `tincd --version` has its own opinion on this.
Let's just get the year from the OS and never touch the message again.

[cet]: https://wiki.ubuntu.com/ToolChain/CompilerFlags#A-fcf-protection
[cfi]: https://lwn.net/Articles/856514/

---

I would have posted this separately, but only cbadb52f011cb8b5a79640ee6bd71d9a2b43f60e can be split off without causing more conflicts (of which are are enough already from previous PRs), and it's only three lines.

### DIFF
--- a/src/have.h
+++ b/src/have.h
@@ -63,6 +63,12 @@
 #endif
 #endif
 
+#if defined(HAVE_ATTR_MALLOC_WITH_ARG)
+#define ATTR_DEALLOCATOR(dealloc) __attribute__((__malloc__(dealloc)))
+#else
+#define ATTR_DEALLOCATOR(dealloc)
+#endif
+
 #ifdef HAVE_ATTR_MALLOC
 #define ATTR_MALLOC __attribute__((__malloc__))
 #else
--- a/src/node.c
+++ b/src/node.c
@@ -99,10 +99,10 @@ void free_node(node_t *n) {
 	sockaddrfree(&n->address);
 
 #ifndef DISABLE_LEGACY
-	cipher_free(&n->incipher);
-	digest_free(&n->indigest);
-	cipher_free(&n->outcipher);
-	digest_free(&n->outdigest);
+	cipher_free(n->incipher);
+	digest_free(n->indigest);
+	cipher_free(n->outcipher);
+	digest_free(n->outdigest);
 #endif
 
 	ecdsa_free(n->ecdsa);
--- a/src/node.h
+++ b/src/node.h
@@ -121,8 +121,8 @@ extern struct node_t *myself;
 extern splay_tree_t node_tree;
 
 extern void exit_nodes(void);
-extern node_t *new_node(void) ATTR_MALLOC;
 extern void free_node(node_t *n);
+extern node_t *new_node(void) ATTR_MALLOC ATTR_DEALLOCATOR(free_node);
 extern void node_add(node_t *n);
 extern void node_del(node_t *n);
 extern node_t *lookup_node(char *name);
--- a/src/splay_tree.h
+++ b/src/splay_tree.h
@@ -63,11 +63,11 @@ typedef struct splay_tree_t {
 
 /* (De)constructors */
 
-extern splay_tree_t *splay_alloc_tree(splay_compare_t compare, splay_action_t delete) ATTR_MALLOC;
 extern void splay_free_tree(splay_tree_t *tree);
+extern splay_tree_t *splay_alloc_tree(splay_compare_t compare, splay_action_t delete) ATTR_MALLOC ATTR_DEALLOCATOR(splay_free_tree);
 
-extern splay_node_t *splay_alloc_node(void) ATTR_MALLOC;
 extern void splay_free_node(splay_tree_t *tree, splay_node_t *node);
+extern splay_node_t *splay_alloc_node(void) ATTR_MALLOC ATTR_DEALLOCATOR(splay_free_node);
 
 /* Insertion and deletion */
 
--- a/meson.build
+++ b/meson.build
@@ -50,6 +50,34 @@ endif
 cc_flags = [cc_defs]
 ld_flags = []
 
+if cc_name != 'msvc'
+  cc_flags += [
+    '-Qunused-arguments',
+    '-Wbad-function-cast',
+    '-Wduplicated-branches',
+    '-Wduplicated-cond',
+    '-Wformat-overflow=2',
+    '-Wformat-truncation=1', # 2 prints too much noise
+    '-Wformat=2',
+    '-Wlogical-op',
+    '-Wmissing-declarations',
+    '-Wmissing-noreturn',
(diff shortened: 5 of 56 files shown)


src/have.h
```diff
@@ -61,6 +61,12 @@
 #endif
 #endif
 
+#if defined(HAVE_ATTR_MALLOC_WITH_ARG)
+#define ATTR_CONSTRUCTOR(dealloc) __attribute__((__malloc__(dealloc)))
```
gsliepen, line 65: The naming is a bit unfortunate because there is also an `__attribute__((constructor))`. I just checked and you can use empty parentheses even with compilers that don't support `__malloc__`-with-arg, so you can make a single macro that covers both cases:
```
#if defined(HAVE_ATTR_MALLOC_WITH_ARG)
#define ATTR_MALLOC(dealloc) __attribute__((__malloc__(dealloc)))
#else
#define ATTR_MALLOC(dealloc) __attribute__((__malloc__))
#endif
```
And then just use empty parentheses when using it on functions that don't have a corresponding deallocator, like so:
```
static inline void *xmalloc(size_t n) ATTR_MALLOC();
```

src/node.c
```diff
@@ -95,10 +95,14 @@ void free_node(node_t *n) {
 	sockaddrfree(&n->address);
 
 #ifndef DISABLE_LEGACY
-	cipher_free(&n->incipher);
-	digest_free(&n->indigest);
-	cipher_free(&n->outcipher);
-	digest_free(&n->outdigest);
+	cipher_free(n->incipher);
+	n->incipher = NULL;
```
gsliepen, line 99: Why is it necessary to NULL these variables if we are going to `free(n)` anyway?

src/version.c
```diff
@@ -28,3 +29,22 @@ const char *const BUILD_VERSION = GIT_DESCRIPTION;
...
+static int current_year(void) {
+	const time_t t = time(NULL);
+	struct tm tm = *localtime(&t);
+	return tm.tm_year + 1900;
+}
+
+extern char *copyright_banner(void) {
+	char *msg = NULL;
+	xasprintf(&msg,
+	          "Copyright (C) 1998-%d Ivo Timmermans, Guus Sliepen and others.\n"
```
gsliepen, line 42: No, this is not how copyright notices are supposed to work. This should be reverted.

A better way would be to have some script analyzing the git log and checking against copyright messages in the source code. In particular, copyright headers at the top of source files should have the year of the last modification of that source file, and the copyright banner we print should have the year of the last modification to anything in src/. But it's better to do this in a different pull request.

src/have.h
hg: I don't think we can do this since they're in effect two completely different attributes (which probably should have been reflected in their names, but I'm not going to argue with ~~gods~~ compiler developers).

afaik `malloc(deallocator)` means "user must call `deallocator` to release resources *returned* by this function" (not *created*, and not necessarily memory); so it's basically a marker to tie pairs of "allocate/deallocate". The PR was posted mostly for this attribute, since there were lots of mismatched calls previously (most were created by myself); let the compiler prevent it in the future.

`malloc` means "this function creates a new island of memory which doesn't have pointers into anything previously existing, thus resembling the behavior of `malloc()`"; it's just an optimization.

The first one can be useful in situations like `get_from_cache()`/`return_to_cache()`, where the second one doesn't apply.

I removed plain `malloc` from functions that return memory allocated by libraries like OpenSSL. Who knows what those opaque structures could contain, or where they could point to. (Also from functions that take a pointer argument, allocate a new struct, and embed that pointer into it, for obvious reasons).

---

https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html

> Attribute malloc indicates that a function is malloc-like, i.e., that the pointer P returned by the function cannot alias any other pointer valid when the function returns, and moreover no pointers to valid objects occur in any storage addressed by P.

> Independently, the form of the attribute with one or two arguments associates deallocator as a suitable deallocation function for pointers returned from the malloc-like function.

> Using the attribute with no arguments is designed to improve optimization by relying on the aliasing property it implies. Functions like malloc and calloc have this property because they return a pointer to uninitialized or zeroed-out, newly obtained storage. However, functions like realloc do not have this property, as they may return pointers to storage containing pointers to existing objects.

> Associating a function with a deallocator helps detect calls to mismatched allocation and deallocation functions ...

> To indicate that an allocation function both satisifies the nonaliasing property and has a deallocator associated with it, both the plain form of the attribute and the one with the deallocator argument must be used.

gsliepen: Ugh, why did they give it the same name... Anyway, good explanation and I agree with the removal of `malloc` from things that are not guaranteed to have malloced exactly the pointer they return.

src/node.h
```diff
@@ -121,8 +121,8 @@ extern struct node_t *myself;
 extern splay_tree_t node_tree;
 
 extern void exit_nodes(void);
-extern node_t *new_node(void) ATTR_MALLOC;
 extern void free_node(node_t *n);
+extern node_t *new_node(void) ATTR_DEALLOCATOR(free_node);
```
gsliepen, line 125: This one should also have `ATTR_MALLOC`

src/splay_tree.h
```diff
@@ -63,11 +63,11 @@ typedef struct splay_tree_t {
 
 /* (De)constructors */
 
-extern splay_tree_t *splay_alloc_tree(splay_compare_t compare, splay_action_t delete) ATTR_MALLOC;
 extern void splay_free_tree(splay_tree_t *tree);
+extern splay_tree_t *splay_alloc_tree(splay_compare_t compare, splay_action_t delete) ATTR_DEALLOCATOR(splay_free_tree);
```
gsliepen, line 67: Keep `ATTR_MALLOC`

hg: Good catch, thanks.

PR #376, opened 2022-04-22
Add building quickstart to the top of README.md

hg:
Written to be as short as possible. We have `INSTALL.md`, but its existence is not very obvious, and it's pretty detailed and long-winded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,8 @@
 # tinc
 
-This is the README file for tinc version 1.1pre18. Installation
-instructions may be found in the INSTALL file.
+This is the README file for tinc version 1.1pre18. Installation instructions may be found in the [INSTALL](INSTALL.md) file.
+
+## Copyright
 
 tinc is Copyright © 1998-2021 Ivo Timmermans, Guus Sliepen <guus@tinc-vpn.org>, and others.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,7 +1,32 @@
 # tinc
 
-This is the README file for tinc version 1.1pre18. Installation
-instructions may be found in the INSTALL file.
+This is the README file for tinc version 1.1pre18. Installation instructions may be found in the [INSTALL](INSTALL.md) file.
+
+## Quickstart
```
gsliepen, line 5: If I see "Quickstart" my first guess and hope would be to see how to quickly start *using* the VPN, not how to compile it. I think this section should just be removed, just adding a real link to the INSTALL.md file at the top is already very helpful. I think I'll rewrite the whole README.md; it's just completely wrong in many ways, and instead it should just be a short file that explains this is an open source mesh VPN with certain features.

gsliepen (commented):
The updates to INSTALL.md look very fine.

README.md
hg: Right, that makes sense. Using `build` as a name for the build directory turned out to be a bad idea since it's really not obvious that it's not a command, so `INSTALL.md` changes are probably worth keeping.

PR #377, opened 2022-04-23
Timestamps and a bit of color in tincd `stderr` logs

hg:
I've been using this recently to make important messages easier to see during debugging. Don't know what your opinion is on toys like this, so just throwing it out there.

Should work on Windows [since ~2015][win] (on older versions [`SetConsoleMode()`][scm] will fail and we keep colors off).

The test on other operating systems is similar to what [GCC does][gcc].

Only the 16-color palette is used, with the exception of gray (it's from the 8-bit one). Should be available everywhere by now.

[win]: https://github.com/chalk/chalk/blob/b17d862944f9efea1ee12aaf80fe19ed5063a34d/source/vendor/supports-color/index.js#L93-L95

[gcc]: https://github.com/gcc-mirror/gcc/blob/16e2427f50c208dfe07d07f18009969502c25dc8/gcc/diagnostic-color.c#L203-L222

[scm]: https://docs.microsoft.com/en-us/windows/console/setconsolemode

image

### DIFF
--- a/src/console.c
+++ b/src/console.c
@@ -1,0 +1,28 @@
+#include "system.h"
+
+#include "console.h"
+
+bool use_ansi_escapes(FILE *out) {
+	bool is_tty = isatty(fileno(out));
+
+#ifndef HAVE_WINDOWS
+	const char *term = getenv("TERM");
+	return is_tty && term && strcmp(term, "dumb");
+#else
+	HANDLE console;
+
+	if(out == stdout) {
+		console = GetStdHandle(STD_OUTPUT_HANDLE);
+	} else if(out == stderr) {
+		console = GetStdHandle(STD_ERROR_HANDLE);
+	} else {
+		return false;
+	}
+
+	DWORD mode = 0;
+	return is_tty &&
+	       console != INVALID_HANDLE_VALUE &&
+	       GetConsoleMode(console, &mode) &&
+	       SetConsoleMode(console, mode | ENABLE_VIRTUAL_TERMINAL_PROCESSING);
+#endif
+}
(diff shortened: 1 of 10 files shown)


hg:
Well, this is weird. The check that's failing UBSAN is disabled:

https://github.com/gsliepen/tinc/blob/1.1/.ci/sanitizers/run.sh#L11

and it was working fine yesterday, but now it stopped caring for some reason.

I think I'll just move sanitizer jobs to containers, similar to how other tests are run. GitHub virtual environments are painful to deal with directly.

gsliepen:
Sure. The only drawback is that you are only coloring the `stderr` output when running tinc in the foreground. It would be nice if the output of `tinc log` and the log messages sent via the umbilical fd when `tinc start` spawns `tincd` would also have the same format and coloring.

hg:
Good idea, thanks for the suggestion.

It's a bit contrived since we need to support multiple concurrent `tinc log`s, some of which may not be able to show colors.

src/console.c
```diff
@@ -0,0 +1,17 @@
+#include "system.h"
+
+#include "console.h"
+
+bool use_ansi_escapes(void) {
+#ifndef HAVE_WINDOWS
+	const char *term = getenv("TERM");
+	return term && strcmp(term, "dumb") && isatty(STDERR_FILENO);
```
gsliepen, line 8: If you run the `tinc log` command it's going to output the log to `stdout`, so then you should check `STDOUT_FILENO` instead. Maybe pass the desired `FILE *` (`stdout`/`stderr`) to this function as a parameter?

PR #378, opened 2022-04-23
Fix a few bugs in proxy support; add tests

hg:
I started writing this to check that exec proxy was not broken by sandboxing, but it's now large enough to post separately.

socks5 authentication didn't actually work since if you request it, socks5 servers send an [additional two-byte message][wiki] with authentication results, which wasn't checked correctly.

[wiki]: https://en.wikipedia.org/wiki/SOCKS#SOCKS5

This not only shows with the hand-rolled Python proxy, but also with `sockd` (which I use heavily with many programs that support authenticated socks5).

Compare proxy responses for anonymous and authenticated socks5:

    anon (ok)   50 50
    anon (err)  50 52
    auth (ok)   52 10 50

Error conditions are not covered by tests yet because tincd bails out too early and doesn't log the messages we need. It treats invalid responses as it should, so it's mostly a cosmetic issue. I'll fix that separately after more important things are done.


src/meta.c
```diff
@@ -287,25 +368,7 @@ bool receive_meta(connection_t *c) {
...
-						if(tcpbuffer[2] != 5) {
-							logger(DEBUG_CONNECTIONS, LOG_ERR, "Invalid response from proxy server");
-							return false;
-						}
-
-						if(tcpbuffer[3] == 0) {
-							logger(DEBUG_CONNECTIONS, LOG_DEBUG, "Proxy request granted");
-						} else {
-							logger(DEBUG_CONNECTIONS, LOG_DEBUG, "Proxy request rejected");
+						if(!parse_socks5((socks5_response_t *) tcpbuffer)) {
```
gsliepen, line 371: While you're at it, it looks like the original code didn't do any length checks. There's no issue with out of bounds access, but it might be good to check that we actually read a whole response packet instead of potentially parsing stale data.

gsliepen (commented):
Apart from that it looks good to me.

hg:
The patch is on the heavy side, but most of it are tests. Those are easy to get rid of.

test/unit/test_proxy.c
```diff
@@ -0,0 +1,430 @@
...
+	sockaddr_t sa = {.sa.sa_family = AF_UNKNOWN};
+	assert_int_equal(0, socks_req_len(PROXY_SOCKS4, &sa));
+	assert_int_equal(0, socks_req_len(PROXY_SOCKS5, &sa));
+}
+
+static void test_check_socks_resp_wrong_types(void **state) {
+	(void)state;
+
+	uint8_t buf[512];
+	assert_false(check_socks_resp(PROXY_NONE, buf, sizeof(buf)));
```
gsliepen, line 111: GCC 11.2.0 is correctly complaining that `buf[]` is used uninitialized here.

hg: Oops, somehow I missed that. Thanks.

PR #379, opened 2022-04-25
Add basic sandboxing for OpenBSD

hg:
Uses `pledge()`/`unveil()` to limit what tincd/tinc can do after initialization finishes.

Paths are only restricted for tincd. It's not difficult to write a similar list for tincctl, but since it's a oneshot program which isn't running for months accepting random connections from around the internet, I don't think it makes much sense to do so, taking support costs into consideration.

`sandbox` is a meson option (and not a boolean) since it may come in handy when adding support for other operating systems (for example, only if libseccomp is present).

Can't call myself an experienced OpenBSD user, so if anyone knowledgeable is willing to test this — be my guest. I've been running this code for a few days without any issues, but it surely has the potential to cause breakage.

---

This PR also makes scripts optional by adding a new configuration variable `DisableScripts`.

Personally, I have no use for them. Network interfaces can be configured through other means.

This is what disabling scripts allows us to do:

- drop access to `fork()` and `execve()` (but only if `proxytype` is not `PROXY_EXEC` — we notify the user in that case)
    - since we cannot put any restrictions on child processes (because they're expected to be able to do anything the user can do), this leaves a giant escape hatch in case of attack.
    - privileges are limited once everything is initialized, so `daemon()` is not a problem.
- even if that somehow happens, limit our children's privileges to `exit()` and nothing else.
- remove access to `scriptinterpreter`, `/bin`, `/sbin`, `/usr/bin`, `/usr/local/bin`, etc.
- make stuff in tinc configuration directory non-executable (without changing file flags).

Here are promises with `DisableScripts = yes`:

```
openbsd$ ps -O pledge | grep [t]incd
20571 stdio,rpath,wpath,cpath,inet,unix,dns    p2  S+pU     0:00.01 ./build/src/tincd -c /tmp/foo -D
```

and without:

```
openbsd$ ps -O pledge | grep [t]incd
65412 stdio,rpath,wpath,cpath,inet,unix,dns,proc,exec    p2  S+pU     0:00.01 ./build/src/tincd -c /tmp/foo -D
```

and here's the [full list][pledge].

[pledge]: https://man.openbsd.org/pledge.2

I don't think we can do much better without splitting tincd into multiple restricted processes like OpenBSD does with its daemons.


hg:
Now, about doing the same for Linux: it's probably going to use `seccomp-bpf` (maybe through libseccomp, but likely without it since rules are likely to be static, and it's [not difficult][sec] to write a BPF program for a static rule list).

[sec]: https://outflux.net/teach-seccomp/

The question is what approach to use: either allow or deny by default.

Obviously, it would be much better to only allow syscalls we actually use, but I worry about third-party libraries. For example, tinc didn't use `getrandom()` until recently, but OpenSSL did, and any library could start using any newfangled syscall when some user builds tinc with a newer version we haven't tested yet.

So I lean towards blocking:

- more dangerous syscalls (like `fork`/`clone`/`execve` if scripts are disabled)
- `pivot_root` since chroot is done before dropping privileges
- and stuff like `create_module`

Another way would be implement both and replace `--no-sandbox` with a more flexible option (like `--sandbox=none`, or `--sandbox=lax`, or `--sandbox=strong`, with the default being `lax`), and let the user opt-in into stricter filtering if they wish to.

src/script.c
```diff
@@ -141,6 +143,11 @@ void environment_exit(environment_t *env) {
 }
 
 bool execute_script(const char *name, environment_t *env) {
+	if(!enable_scripts) {
+		logger(DEBUG_ALWAYS, LOG_NOTICE, "Not running script %s as scripts are disabled", name);
```
gsliepen, line 147: We shouldn't warn about this, if `enable_scripts` is `false`, it's because the user explicitly disabled it.

src/script.c
```diff
@@ -141,6 +143,11 @@ void environment_exit(environment_t *env) {
 }
 
 bool execute_script(const char *name, environment_t *env) {
+	if(!enable_scripts) {
```
gsliepen, line 146: We might still want to allow `tinc-up` to be executed; it's run before privileges are dropped. This might make it easier for people who are used to the `tinc-up` script to configure the network.

src/proxy.c
```diff
@@ -283,3 +284,22 @@ size_t create_socks_req(proxytype_t type, void *buf, const sockaddr_t *sa) {
...
+char *exec_proxy_path(void) {
+	if(proxytype == PROXY_EXEC && proxyhost && *proxyhost) {
+		char *command = xstrdup(proxyhost);
+
+		// Find the first non-whitespace token
+		const char *exe = strtok(command, " \t\r\n");
+
+		if(exe) { // If found, remove whitespace prefix and return the path
+			memmove(command, exe, strlen(exe) + 1);
+			return command;
```
gsliepen, line 297: But `command` might not necessarily be a path to an executable, it can be an arbitrary command line which is passed to `system()`.

src/script.c
hg: Hmm. I'm not sure how to reconcile that with the new option. If a user disables script execution, she expects them not to run, no?

Maybe we could remove the option altogether, and add another one that configures the isolation level. Such as:

`Sandbox = off | medium | strong`

- `off`: tincd continues working as it does now; if `medium` is implemented properly, this shouldn't be needed by anyone.
- `medium`: used by default; on OpenBSD it would probably be a call to `pledge()` and nothing else.
- `strong`: most scripts don't work, with the exception of `tinc-up`; exec proxies won't run; probably some other limitation I can't think of right now.

gsliepen: That sounds great!

hg:
I think this is about as restrictive as I can make it.

Non-standard executable paths are not available even on `normal` level since doing that pretty much requires `rx` access to everything (including home directories and other interesting places).

We could add another level (`low`) which permits that, but it's probably better to encourage users to put binaries into `/usr/local/bin` and be done with it.

src/meson.build
```diff
@@ -260,7 +260,7 @@ if not opt_miniupnpc.disabled()
   endif
 endif
 
-if not opt_curses.disabled()
+if opt_curses.enabled() or (opt_curses.auto() and os_name != 'windows')
```
gsliepen, line 263: Why disable it by default for Windows? This also doesn't seem related to OpenBSD sandboxing at all.

src/net_setup.c
```diff
@@ -231,10 +232,13 @@ char *get_name(void) {
...
-	free(scriptinterpreter);
-	scriptinterpreter = NULL;
-
-	get_config_string(lookup_config(&config_tree, "ScriptsInterpreter"), &scriptinterpreter);
+	if(sandbox_can(START_PROCESSES, AFTER_SANDBOX) && sandbox_can(USE_NEW_PATHS, AFTER_SANDBOX)) {
+		free(scriptinterpreter);
+		scriptinterpreter = NULL;
+		get_config_string(lookup_config(&config_tree, "ScriptsInterpreter"), &scriptinterpreter);
+	} else {
+		logger(DEBUG_ALWAYS, LOG_NOTICE, "Not reloading ScriptsInterpreter because of sandbox.");
```
gsliepen, line 240: A bit annoying if there was no `ScriptsInterpreter` set to begin with, or if it wasn't changed.

src/tincd.c
```diff
@@ -322,6 +323,27 @@ static bool parse_options(int argc, char **argv) {
...
+
+	if(get_config_string(lookup_config(&config_tree, "Sandbox"), &value)) {
+		if(!strcasecmp("off", value)) {
+			level = SANDBOX_NONE;
+		} else if(!strcasecmp("normal", value)) {
+			level = SANDBOX_NORMAL;
+		} else if(!strcasecmp("high", value)) {
+			level = SANDBOX_HIGH;
+		} else {
+			logger(DEBUG_ALWAYS, LOG_ERR, "Bad sandbox value %s, using default", value);
```
gsliepen, line 338: That's bad. Consider making a typo, thinking Sandbox level is high, but tinc just using normal level. Yes there is a log message, but it's easy to miss. Another issue is if Sandbox level is set explicitly on a platform that we don't have sandboxing implemented for. It would be safer to exit tinc immediately if an unsupported or unknown value for Sandbox is used.

gsliepen (commented):
I tested it out on OpenBSD with some realistic configurations, and it seems to work as expected, except when making typos in the `Sandbox` configuration option.

src/meson.build
hg: To have fully green CI here.

- https://github.com/gsliepen/tinc/runs/6247115066#step:8:355
- https://github.com/gsliepen/tinc/runs/6248463823#step:8:335

Broken by a recent change to the `ncurses` package. Probably [this][commit] one. I'll look into it later.

[commit]: https://github.com/msys2/MINGW-packages/pull/11586

hg:
Thanks for the review. I think it would be best to merge #382 first. libgcrypt support was broken and I nearly missed that because it's painful to continuously retest all configurations manually.

It's caused by 'secure memory', which is initialized using `mlock()` on first allocation (unless you do it separately), which fails since we've dropped privileges by that point.

gsliepen:
> It's caused by 'secure memory', which is initialized using `mlock()` on first allocation (unless you do it separately), which fails since we've dropped privileges by that point.

Good catch!

PR #382, opened 2022-05-01
CI: run all test flavors on BSDs

hg:
Let's try running all test flavors on BSDs too.

We're not using the same scripts here since sourcehut does not support saving build artifacts on failure, and scripts are too different anyway.

Also use system versions of OpenSSL/curses/readline.


hg:
The failure in mingw is caused by a recent change to the ncurses package and is fixed by #379.

hg:
I additionally replaced `getrandom()` with `getentropy()`, which is present on most of the supported operating systems:

- [Linux](https://man7.org/linux/man-pages/man3/getentropy.3.html)
- [macOS](https://keith.github.io/xcode-man-pages/getentropy.2.html)
- [OpenBSD](https://man.openbsd.org/getentropy)
- [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=getentropy&sektion=3&format=html)
- [Illumos](https://illumos.org/man/3C/getentropy)
- [DragonflyBSD](https://man.dragonflybsd.org/?command=getentropy&section=3)

It should have been used from the start if I were more of a BSD guy.

Since Windows has its own `CryptGenRandom()`, `/dev/urandom` is now only used by NetBSD.

PR #386, opened 2022-05-15
Add Linux sandbox (seccomp-bpf + Landlock LSM)

hg:
Landlock is similar to `unveil()` and is available on Linux 5.13+.

- https://lwn.net/Articles/859908
- https://docs.kernel.org/security/landlock.html

Since both are inherited by child processes, and there's no way to disable this, we use a trick similar to what browsers are doing — fork a process early, drop privileges on the main one, and use the privileged process for running scripts.

This also lets us remove access to `fork()`/`execve()` on OpenBSD for both sandbox levels.

# Hardware

Only amd64 and aarch64 are supported because I don't have access to anything else, and it's pretty dangerous to "support" other architectures without actually testing them. For example, some architectures use `socketcall` instead of separate `accept`/`bind`/`connect`/etc, some implement `gettimeofday` through [VDSO][vdso] instead of an actual syscall, and so on.

[vdso]: https://lwn.net/Articles/795128

If a user wishes to go ahead anyway, they can force-enable the sandbox with:

    $ meson setup build -D sandbox=enabled

# Security

There is at least one way to circumvent the sandbox, but I believe it will only work with a non-empty `ScriptsInterpreter`. Since tincd has full access to the `hosts` subdirectory, the attacker can create a `hosts/xxx-up` or `hosts/xxx-down` script and ask the script worker to execute it.

It shouldn't be possible with an empty interpreter since you need to make the script executable, and both `umask()` and all `chmod`-related syscalls are blocked by seccomp.

Additionally, we remove write access to existing scripts inside `hosts/` to prevent broken tincd from rewriting them and gaining shell access.

Reassigning `ScriptsInterpreter` to another value at runtime shouldn't be possible since script worker uses its own copies of all configuration variables, and access to other processes' memory is prevented by seccomp.

tincd also doesn't have write access to its own configuration, so it cannot rewrite the config and restart itself.

# Performance

The PR doesn't seem to affect simple `iperf3` between two nodes in any way. The results may be different with hundreds of nodes, but I don't have the hardware to test this.

## With seccomp-bpf and Landlock

```
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  5]   0.00-1.00   sec  99.4 MBytes   834 Mbits/sec    0   3.15 MBytes
[  5]   1.00-2.00   sec  93.8 MBytes   786 Mbits/sec    0   3.15 MBytes
[  5]   2.00-3.00   sec  95.0 MBytes   797 Mbits/sec  4112   1.19 MBytes
[  5]   3.00-4.00   sec  98.8 MBytes   828 Mbits/sec    0   1.30 MBytes
[  5]   4.00-5.00   sec  98.8 MBytes   828 Mbits/sec   10   1.38 MBytes
[  5]   5.00-6.00   sec   115 MBytes   965 Mbits/sec  653    631 KBytes
[  5]   6.00-7.00   sec   114 MBytes   954 Mbits/sec    1    567 KBytes
[  5]   7.00-8.00   sec   114 MBytes   954 Mbits/sec    1    501 KBytes
[  5]   8.00-9.00   sec   115 MBytes   965 Mbits/sec    0    652 KBytes
[  5]   9.00-10.00  sec   114 MBytes   954 Mbits/sec    5    591 KBytes
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec  1.03 GBytes   887 Mbits/sec  4782             sender
[  5]   0.00-10.01  sec  1.03 GBytes   884 Mbits/sec                  receiver
```

## Without sandbox

```
[  5] local 192.168.1.1 port 57292 connected to 192.168.1.2 port 5201
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  5]   0.00-1.00   sec   107 MBytes   899 Mbits/sec  222    420 KBytes
[  5]   1.00-2.00   sec  98.8 MBytes   828 Mbits/sec    0    574 KBytes
[  5]   2.00-3.00   sec   104 MBytes   870 Mbits/sec    0    701 KBytes
[  5]   3.00-4.00   sec   106 MBytes   891 Mbits/sec    7    634 KBytes
[  5]   4.00-5.00   sec   100 MBytes   839 Mbits/sec    3    551 KBytes
[  5]   5.00-6.00   sec   106 MBytes   891 Mbits/sec    0    686 KBytes
[  5]   6.00-7.00   sec   108 MBytes   902 Mbits/sec    4    619 KBytes
[  5]   7.00-8.00   sec   108 MBytes   902 Mbits/sec    1    547 KBytes
[  5]   8.00-9.00   sec   109 MBytes   912 Mbits/sec    0    684 KBytes
[  5]   9.00-10.00  sec   108 MBytes   902 Mbits/sec    4    621 KBytes
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec  1.03 GBytes   884 Mbits/sec  241             sender
[  5]   0.00-10.01  sec  1.03 GBytes   881 Mbits/sec                  receiver
```


src/script.c
```diff
@@ -141,15 +145,234 @@ void environment_exit(environment_t *env) {
...
+#ifdef HAVE_SANDBOX
+
+	if(!sandbox_can(RUN_SCRIPTS, RIGHT_NOW)) {
 		return false;
 	}
 
+	if(worker.stopped) {
+		logger(DEBUG_ALWAYS, LOG_ERR, "Script worker is broken, scripts will not work until tincd is restarted");
+		return false;
+	}
```
gsliepen, line 366: I wonder what we should do here. It's one thing that a script itself fails, but if the script worker fails we won't even try to run the scripts themselves anymore. If someone is relying on something in a script, like say setting firewall rules to allow or block traffic from a node that just came online, this will be problematic. Perhaps it's better to just terminate tincd altogether when that happens.

src/script.c
```diff
@@ -141,15 +145,234 @@ void environment_exit(environment_t *env) {
...
+
+static bool read_unix(int fd, void *data, size_t len) {
+	ssize_t res = read(fd, data, len);
+	return res > 0 && (size_t)res == len;
+}
+
+static bool write_unix(int fd, const void *data, size_t len) {
+	ssize_t res = write(fd, data, len);
+	return res > 0 && (size_t)res == len;
+}
```
gsliepen, line 238: Most UNIX socket implementations are always reliable and in-order, whether it is stream or datagram oriented, but what is not guaranteed on a `SOCK_STREAM` socket is that a `write()` will complete in its entirety, nor that every `read()` will return exactly what a single `write()` sent. Either use `SOCK_DGRAM`, or even better, `SOCK_SEQPACKET`. You then also don't have to encode the length in a header.

hg: FWIW, `SEQPACKET` is not available on NetBSD and macOS. NetBSD doesn't seem to have any sandboxing capabilities, but macOS does. I'm not planning on implementing it since I don't have access to macOS and don't care for it much, but if someone does, that's something to think about.

It's a lot nicer than sending lengths manually, so I left `SEQPACKET` for now.

src/script.c
hg: Good idea, this gives the supervisor (like systemd) an opportunity to 'fix' the issue.

src/sandbox.c
```diff
@@ -14,6 +14,14 @@ void sandbox_set_level(sandbox_level_t level) {
...
+bool sandbox_enabled(void) {
+	return false;
+}
+
+bool sandbox_active(void) {
+	return false;
+}
+
 bool sandbox_enter(void) {
 	return true;
```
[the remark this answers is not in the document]
hg: I don't think so. The result is only used to determine if initialization failed and we should terminate the process. It would be more obvious with a proper type system. I can at least add a comment.

hg:
Thanks for the review, fixed. No need to rush this, the change is quite intrusive and risky.

hg:
I'll mark this as WIP for now, best to merge this after everything else is in.

hg:
After some recent changes to `debian:testing`, it fails to even resolve `localhost` without access to NSS dynamic libraries (blocking access to `nsswitch.conf` doesn't help), so I had to open a lot more paths (luckily it doesn't need write access anywhere).

I'll work on test coverage a bit more before marking this as finished.

Issue #373, opened 2022-04-20
Improve the recently seen address cache.


gsliepen:
Tinc supports storing the recently seen addresses of other nodes in an on-disk cache, so that even if their address doesn't match that in their host config files (or if their host config files don't contain `Address`es at all), it can still reconnect to them. There are some issues though:

1. The address cache is not enabled by default. Ensure the `cache/` directory is created at init time.
2. It seems we cache both UDP and TCP sockaddrs, but it probably only makes sense to store TCP ones.
3. We should update the cache in more situations, like after a succesful join using an invitation, and when we update the graph and see a node becomes reachable.
4. Integration tests should be added that check this functionality.

PR #388, opened 2022-05-20
Improve recently seen address cache

hg:
See #373.

Cache directory is recreated as necessary. While it may be annoying for some users, I don't think it has any real downsides (besides maybe adding a small amount of flash wear on old routers), and there are lots of existing configurations out there that could benefit from this cache. I don't expect many users to read release notes for all software they use.

### DIFF
--- a/lint.py
+++ b/lint.py
@@ -39,6 +39,7 @@ linters = (
     ["pylint", "."],
     ["mypy", "--exclude", "build", "."],
     ["shellcheck", "-x", *glob(".ci/**/*.sh", recursive=True)],
+    ["markflow", "--line-length", "80", "--check" if DRY else "--verbose", ".", ".ci"],
 )
 
 failed: bool = False
--- a/src/fs.c
+++ b/src/fs.c
@@ -1,0 +1,142 @@
+#include "system.h"
+
+#include "fs.h"
+#include "names.h"
+#include "xalloc.h"
+
+static bool makedir(const char *path, mode_t mode) {
+	assert(path);
+
+	if(mkdir(path, mode)) {
+		if(errno == EEXIST) {
+			chmod(path, mode);
+			return true;
+		}
+
+		fprintf(stderr, "Could not create directory %s: %s\n", path, strerror(errno));
+		return false;
+	}
+
+	return true;
+}
+
+bool makedirs(tinc_dir_t dirs) {
+	if(!dirs) {
+		return false;
+	}
+
+	char path[PATH_MAX];
+	bool need_confbase = dirs & ~((unsigned)DIR_CONFDIR);
+
+	if(need_confbase && (!confbase || !makedir(confbase, 0755))) {
+		return false;
+	}
+
+	if(dirs & DIR_CONFDIR && !confbase_given && (!confdir || !makedir(confdir, 0755))) {
+		return false;
+	}
+
+	if(dirs & DIR_INVITATIONS) {
+		snprintf(path, sizeof(path), "%s" SLASH "invitations", confbase);
+
+		if(!makedir(path, 0700)) {
+			return false;
+		}
+	}
+
+	if(dirs & DIR_CACHE) {
+		snprintf(path, sizeof(path), "%s" SLASH "cache", confbase);
+
+		if(!makedir(path, 0755)) {
+			return false;
+		}
+	}
+
+	if(dirs & DIR_HOSTS) {
+		snprintf(path, sizeof(path), "%s" SLASH "hosts", confbase);
+
+		if(!makedir(path, 0755)) {
+			return false;
+		}
+	}
+
+	return true;
+}
+
+
+/* Open a file with the desired permissions, minus the umask.
+   Also, if we want to create an executable file, we call fchmod()
+   to set the executable bits. */
+
+FILE *fopenmask(const char *filename, const char *mode, mode_t perms) {
+	mode_t mask = umask(0);
+	perms &= ~mask;
+	umask(~perms & 0777);
+	FILE *f = fopen(filename, mode);
+
+	if(!f) {
+		fprintf(stderr, "Could not open %s: %s\n", filename, strerror(errno));
+		return NULL;
+	}
+
+#ifdef HAVE_FCHMOD
+
+	if(perms & 0444) {
+		fchmod(fileno(f), perms);
+	}
+
+#endif
+	umask(mask);
+	return f;
+}
+
+char *absolute_path(const char *path) {
+#ifdef HAVE_WINDOWS
+	// Works for nonexistent paths
+	return _fullpath(NULL, path, 0);
+#else
+
+	if(!path || !*path) {
+		return NULL;
+	}
+
+	// If an absolute path was passed, return its copy
+	if(*path == '/') {
+		return xstrdup(path);
+	}
+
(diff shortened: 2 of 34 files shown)


hg:
This also flushed out an old bug which I've often run into in the past:

1. setup node `pub` with a publicly available IP and port.
1. put node `nat` behind NAT or firewall.
1. join them somehow so they know each other's addresses.
1. `pub` starts connecting to `nat` and fails because of firewall/NAT.
1. meanwhile, `nat` successfully connects back to `pub`.
1. `pub` still thinks it has a pending outgoing connection to `nat`.
1. `pub` adds `nat`'s public address to address cache (in `pong` handler, `graph.c`, or any other place), not realizing it's using a "client" port which won't be available anymore after the current connection is closed.

One might wonder why node `nat` is being given a public address when it cannot accept incoming connections, but it makes sense if there are multiple nodes behind the firewall. External nodes can be configured to ignore them, at the cost of much additional complexity in Ansible playbooks.

Here's how it looks in logs:

    2022-05-20 18:12:13 NOTICE  Connection from x.y.z.q port 60310
    2022-05-20 18:12:13 NOTICE  Trying to re-establish outgoing connection in 5 seconds
    2022-05-20 18:12:13 NOTICE  Connection with foo (x.y.z.q port 60310) activated
    2022-05-20 18:12:18 INFO    Already connected to foo
    2022-05-20 18:12:18 DEBUG   Caching recent address for foo

After restarting tincd it uses wrongly cached ports (both nodes should use 655):

    2022-05-20 18:14:56 DEBUG   Error while connecting to foo (x.q.z.y port 35870): Connection refused
    2022-05-20 18:14:57 DEBUG   Error while connecting to bar (x.y.z.q port 60310): Connection refused

The last commit attempts to fix that. It may be completely wrong because I'm not sure why we're retrying outgoing connections if there already is a working connection.

src/fs.c
```diff
@@ -0,0 +1,144 @@
...
+	if(mkdir(path, mode) && errno != EEXIST) {
+		fprintf(stderr, "Could not create directory %s: %s\n", path, strerror(errno));
+		return false;
+	}
+
+	return true;
+}
+
+bool makedirs(tinc_dir_t dirs) {
+#define subdir(sub) snprintf(path, sizeof(path), "%s" SLASH sub, confbase)
```
gsliepen, line 23: I'd rather avoid macros.

We do call `chdir(confbase)` very early when tinc starts up, so perhaps we can just simply do `makedir("sub")`.

test/integration/address_cache.py
```diff
@@ -0,0 +1,79 @@
...
+    log.info("remove %s cache directory", bar)
+    shutil.rmtree(bar.sub("cache"))
+
+    log.info("start node %s", bar)
+    bar.cmd("start")
+    bar[foo.script_up].wait()
+
+    log.info("make sure %s cache was recreated", bar)
+    check.file_exists(bar.sub(f"cache/{foo}"))
+
```
gsliepen, line 76: It might also be nice to actually check that the address cache contains useful information, by removing stopping the nodes, removing `Address`es from the all host config files, and then seeing if they can reconnect just from the information in the cache.

src/net_packet.c
```diff
@@ -193,15 +193,7 @@ static void udp_probe_h(node_t *n, vpn_packet_t *packet, length_t len) {
...
 	   packet used. */
-	if(!n->status.udp_confirmed) {
-		n->status.udp_confirmed = true;
-
-		if(!n->address_cache) {
-			n->address_cache = open_address_cache(n);
-		}
-
-		reset_address_cache(n->address_cache, &n->address);
-	}
```
gsliepen, line 204: But sometimes we only learn the correct address via UDP. For example, nodes A, B and C, A connects to B, B connects to C, but B doesn't have the correct address of A (maybe due to NAT somewhere), but A learns the correct address from C, then A sends C UDP packets, without NAT being in the way, and without initiating a meta-connection (because the autoconnect algorithm might already be satisfied somehow).

src/fs.c
hg: Most calls to this function are done by tincctl, not tincd. The macro actually doesn't save that much, let's remove it.

src/net_packet.c
hg: Sure, but `n->address` is a UDP address, and it means possibly pushing out cached TCP addresses. `n->connection->address` is not guaranteed to be initialized here.

I'm testing all this on a 15-node network with multiple firewalls and layers of NAT in front of some nodes.

If this change is reverted to the original state, tincd starts replacing perfectly working TCP ports that were cached in other places with other ports that look weird (like 772) and don't accept incoming connections. After restarting tincd you get a bunch of "Connection refused" from nodes that could be connected to before.

I'm probably missing something obvious here. I'll keep this change for now as it works better without this code, and dig more into tinc internals. Everything else should be fixed.

test/integration/address_cache.py
hg: Makes sense, added.

src/net_packet.c
gsliepen: But we should know the TCP port `n` is listening on, as that is in the edge information. But yes, we can keep this change and revisit this in another ticket.

gsliepen:
I'm getting a failure in the fs test:
```
not ok 10 - test_fopenmask_new
# 0x1e8 != 0x1c0
# ../test/unit/test_fs.c:210: error: Failure!
```
The reason is that I have my umask set to 077. The fs unit tests should probably call `umask(0)` before doing anything else.

src/net_packet.c
hg: Right, the edge address always seems to be valid in my testing.

gsliepen (commented):
Thanks for adding MarkFlow to the linter. It had some [issue breaking long links](https://github.com/duo-labs/markflow/issues/64), I worked around it by changing one URL. Then again, these files are not changed that often so it's probably fine.

lint.py
```diff
@@ -39,6 +39,7 @@
     ["pylint", "."],
     ["mypy", "--exclude", "build", "."],
     ["shellcheck", "-x", *glob(".ci/**/*.sh", recursive=True)],
+    ["markflow", "--line-length", "80", "--check" if DRY else "--verbose", "."],
```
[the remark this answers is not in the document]
hg: Thanks for noticing. I tried to do without hardcoding paths, but `./.*` grabs too much (including `..`), and I don't want to use globs. Hopefully there will be no more hidden directories.

> worked around it by changing one URL

I think that particular URL could also be moved to a footnote:

```markdown
Make sure the [IP range][cidr] you are choosing is large enough for all the
...

[cidr]: https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing
```

Doesn't matter in this case, but might be useful when there's no short version of the same URL.

gsliepen:
Thanks, nice work again! One thing I want to note though is that you often add unrelated commits to a pull request. For example, in this PR you fix an outgoing connection issue and add Markdown reformatting to lint.py. Those have nothing to do with the address cache. The refactoring of filesystem-related functions is somewhat related, as it cleans up some of the code dealing with making directories, but it could also have been in a separate PR. Splitting them makes reviewing easier and increases the likelihood that I can merge individual PRs earlier.

PR #395, opened 2022-05-23
CI: use compilation database for clang-tidy job

hg:
I didn't realize LLVM already ships a [runner][run] for `clang-tidy` which picks files from the compilation database. This lets us avoid wrapping optional C files in giant `#ifdef`s.

[run]: https://clang.llvm.org/extra/doxygen/run-clang-tidy_8py_source.html

By the way, we're still only running a single check. I'll start adding more one by one and fixing whatever is reported while working on other things.


hg:
Here are some checks for starters. A few suggested issues are actually interesting (`printf` in signal handler, zero `alloca()`, a couple of memory leaks).

Filling `sockaddr_ll` in raw socket device code with `'0'` (as opposed to `0`) looks like a typo. Other examples I'm reading around the 'net use `0`. Is this correct? Sorry, no idea how to quickly test this device.

src/raw_socket_device.c
```diff
@@ -34,8 +34,8 @@
 static const char *device_info = "raw_socket";
 
 static bool setup_device(void) {
-	struct ifreq ifr;
-	struct sockaddr_ll sa;
+	struct ifreq ifr = {0};
+	struct sockaddr_ll sa = {0};
```
gsliepen, line 38: Hm, this was problematic in the past, with compilers complaining about not enough parentheses around initializers. The `'0'` in the `memset()` was definitely bad though.

src/tincctl.c
```diff
@@ -1519,7 +1519,10 @@ static int cmd_pcap(int argc, char *argv[]) {
 static void sigint_handler(int sig) {
 	(void)sig;
 
-	fprintf(stderr, "\n");
+	if(write(2, "\n", 2) < 0) {
```
gsliepen, line 1522: Should be 1, not 2.

src/top.c
```diff
@@ -249,7 +249,7 @@ static void redraw(void) {
 
 	if(changed) {
 		n = 0;
-		sorted = xrealloc(sorted, node_list.count * sizeof(*sorted));
+		sorted = xrealloc(sorted, node_list.count * sizeof(sorted));
```
gsliepen, line 252: The original `sizeof(*sorted)` is correct, sorted itself is just the pointer to the array, we want the size of an array element here.

src/raw_socket_device.c
hg: We have a lot of such statements. I'm counting more than 50, most of them structs. Should I replace this anyway?

```
src/fd_device.c
41:	struct iovec iov = {0};
43:	struct msghdr msg = {0};

src/multicast_device.c
34:static mac_t ignore_src = {0};

src/process.c
48:static SERVICE_STATUS status = {0};

src/protocol.c
222:	past_request_t *new, p = {0};

src/protocol_edge.c
73:	sockaddr_t address, local_address = {0};

src/control.c
155:	sockaddr_t sa = {0};

src/splay_tree.c
28:	splay_node_t left = {0}, right = {0};

src/net.c
491:	signal_t sighup = {0};
492:	signal_t sigterm = {0};
493:	signal_t sigquit = {0};
494:	signal_t sigint = {0};
495:	signal_t sigalrm = {0};

src/ifconfig.c
91:	subnet_t address = {0};
174:	subnet_t subnet = {0}, gateway = {0};

src/node.c
149:	node_t n = {0};

src/linux/device.c
67:	struct ifreq ifr = {0};
119:		struct ifreq ifr_mac = {0};

src/openssl/prf.c
33:	digest_t digest = {0};

src/netutl.c
59:	struct addrinfo *ai, hint = {0};
79:	struct addrinfo *ai, hint = {0};
80:	sockaddr_t result = {0};

src/script.c
54:	static list_t list = {0};

src/route.c
122:	struct ip ip = {0};
123:	struct icmp icmp = {0};
224:	struct icmp6_hdr icmp6 = {0};

src/conf_net.c
5:	subnet_t subnet = {0};

src/sptps_test.c
327:	sockaddr_t sa = {0};

src/net_setup.c
640:	struct addrinfo *ai, hint = {0};

src/tincctl.c
680:	SERVICE_STATUS status = {0};
1851:			subnet_t s = {0};

src/protocol_subnet.c
48:	subnet_t s = {0}, *new, *old;
164:	subnet_t s = {0}, *find;

src/net_packet.c
1004:			node_id_t nullid = {0};
1710:	node_id_t nullid = {0};
1877:	sockaddr_t addr = {0};

test/unit/test_subnet.c
320:		subnet_t sub = {0};
352:	subnet_t sub = {0};
465:	const subnet_t sub = {0};
```

src/tincctl.c
hg: Oh rats, I made a mess of my own. Thanks for the review. FWIW, `STDERR_FILENO` doesn't work here because of msvc.

src/top.c
hg: It's a pointer to a pointer, though? I checked it with `_Static_assert` before pushing just to make sure. Both of these pass:

```c
_Static_assert(sizeof(*sorted) == sizeof(void*), "foo");
_Static_assert(sizeof(nodestats_t) != sizeof(void*), "foo");
```

src/raw_socket_device.c
gsliepen: Well, it passes the pre-existing CI checks, so I'm fine with the `{0}` then :)

src/top.c
gsliepen: Yes, but if the type of `sorted` is ever changed, then `sizeof(sorted)` might become really wrong. Was clang-tidy complaining about this?

gsliepen: I don't like adding comments to appease linters, it quickly becomes a mess otherwise. You already disabled several checks, I would disable this one as well if it generates false positives.

hg: Ah, ok. Makes sense. Although those are disabled temporarily (hopefully) as especially problematic. Disabling one or two rules allows us to enable 10-20 others by globbing the whole group. I expect to fix most of them later.

Issue #338, opened 2021-10-03
Systemd Notify


splitice:
Currently investigating an issue with 1.0 where process is left in "D" (Uninterruptable Sleep) randomly.

Got me thinking that it would be nice if `sd_notify` could be called in the IO loop.

Thoughts?

gsliepen:
Yes, although we don't want to notify systemd every time we call `select()` of course. We can tie it to the periodic checks, like `periodic_handler()` in tinc 1.1 and `check_dead_connections()` in tinc 1.0.

Also, I'd like to avoid needing to link with libsystemd; the protocol is simple enough: check if `$NOTIFY_FD` is set, if so just send `"WATCHDOG=1"` to that fd. We can also use it to notify when tinc has finished initializing and when it's about to quit.

splitice:
Correct, although I'm less keen on re-implementing libsystemd. I'd rather link against the system lib if support is enabled.

Who knows what will change in the future and what hackery people using it might have. Better to use the system supplied library if enabled in my books.

And yes a `timeout_add` for the watchdog makes sense.

PR #396, opened 2022-05-23
Add optional systemd integration

hg:
- startup & shutdown notifications
- optional watchdog with auto-restart on hang (enabled if `WatchdogSec` is set in systemd unit)

re #338

Static analysis job will fail until this is rebased on top of #395.

---

The code is using libsystemd because the library is going to be present on any systemd-based distribution anyway. Otherwise, we're going have to handle things like:

> If the $WATCHDOG_USEC environment variable is set, and the $WATCHDOG_PID variable is unset or set to the PID of the current process, the service manager expects notifications from this process.

> If the usec parameter is non-NULL, sd_watchdog_enabled() will write the timeout in µs for the watchdog logic to it.

(which really means parsing data passed by systemd in environment variables)

> If the first character of $NOTIFY_SOCKET is "@", the string is understood as Linux abstract namespace socket.

and so on.

---

There's also the `RELOADING` event, but I see no point in adding it since configuration reloading is nearly instantaneous.

### DIFF
--- a/src/linux/watchdog.c
+++ b/src/linux/watchdog.c
@@ -1,0 +1,71 @@
+#include "../system.h"
+
+#include <systemd/sd-daemon.h>
+
+#include "../event.h"
+#include "../logger.h"
+#include "../watchdog.h"
+
+static timeout_t timer;
+static struct timeval interval;
+
+static uint64_t second_to_microsecond(time_t second) {
+	return second * 1000000;
+}
+
+static time_t microsecond_to_second(uint64_t micros) {
+	return (time_t)(micros / 1000000);
+}
+
+// Ignore errors from sd_notify() since there's nothing we can do if it breaks anyway.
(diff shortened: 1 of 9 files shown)


src/linux/watchdog.c
```diff
@@ -0,0 +1,66 @@
+#include "../system.h"
+
+#include <systemd/sd-daemon.h>
+
+#include "../event.h"
+#include "../logger.h"
+#include "../watchdog.h"
+
+#define SECOND_TO_MICROSECOND(sec) ((sec) * 1000000)
+#define MICROSECOND_TO_SECOND(us) ((us) / 1000000)
```
gsliepen, line 10: Prefer regular functions over macros.

gsliepen (commented):
Apart from the macros it looks good to me. I'll wait until it's rebased on top of #395 before merging.

PR #398, opened 2022-05-26
Add integration tests for most tincctl commands

hg:
The RHEL 9 clone we're using has just been released. I think these are the usual mirror syncing issues, it should start working soon.

---

Things to do later:

- `REQ_CONNECT` doesn't seem to be implemented at all, `tinc connect` does nothing.
- half of new tests are disabled on Windows because tincctl connections often hang inexplicably. Since tincctl uses UNIX sockets everywhere but Windows, and TCP on Windows, there's probably a bug there. I don't currently have access to that OS, maybe I'll look into that in a few weeks.


.ci/test/run.sh
```diff
@@ -24,7 +24,7 @@ run_tests() {
     sudo pkill -KILL -x "$name" || true
   done
 
-  sudo chown -R "${USER:-$(whoami)}" .
+  sudo chown -R "${USER:-$(whoami)}" . || true
```
hg, line 27: sudo does nothing on mac (it's aliased to a no-op), so tar can (and does) fail with `EPERM`.

gsliepen:
Tinc runs under Wine, I wonder if the issue is reproducable there?

.ci/test/run.sh
hg: I mean it's disabled in CI scripts since sudo isn't needed on macOS and using it actually broke something there (probably homebrew package installation, or something like that). To keep the same scripts working everywhere, we alias sudo to nothing.

https://github.com/gsliepen/tinc/blob/1.1/.ci/test/run.sh#L55-L57

https://github.com/hg/tinc/runs/6604277651?check_suite_focus=true#step:6:12198

hg:
I remember tests having terrible issues running under Wine (both the shell scripts, and the newer Python-based suites). That's why mingw cross-compilation job doesn't run any tests. Let's use this as an opportunity to try to fix both.

hg:
Windows tests shoud work after rebasing on top of #399.

https://github.com/hg/tinc/actions/runs/2396396271

PR #399, opened 2022-05-27
Don't tarpit localhost connections

hg:
The reason Windows tests were hanging was pretty obvious in retrospect — since control connections use TCP there, connection limits apply to them, so they were getting tarpitted.

We could disable tarpit for tests, but I'm pretty sure this is also useful on end-user machines since it's really easy to get blocked by calling tincctl a dozen times in a row fromt a script.


gsliepen:
D'oh!

Issue #397, opened 2022-05-25
Add backwards compatibility tests


gsliepen:
We should ensure 1.1 is backwards compatible with 1.0, and it might also be prudent to check if the latest commit in the 1.1 branch is compatible with the latest 1.1pre* release, so we can avoid surprises.

This should be done in the CI infrastructure. I don't think it's necessary to test this on all platforms, just one Linux platform.

PR #400, opened 2022-05-28
Add compatibility checks to CI (old tinc & muon)

hg:
## old tinc

re #397

Any idea what other useful checks can be added that do not require reimplementing a second test suite?

## muon

Ensures we have some level of compatibility with [muon][muon] to have tinc buildable on systems with only a C compiler.

Unit tests won't run since it [doesn't consider][check] `SIGABRT` to be a 'proper' way of terminating a `must_fail` test.

Integration tests don't work for obvious reasons.

[muon]: https://git.sr.ht/~lattis/muon
[check]: https://git.sr.ht/~lattis/muon/tree/master/item/src/platform/run_cmd.c#L158-163


INSTALL.md
```diff
@@ -1,14 +1,27 @@
...
 
+## Required
+
 Before you can start compiling tinc from a fresh git clone, you have
 to install the very latest versions of the following packages:
 
-- `meson`
-- `ninja`
+- `meson` or `muon` (read below)
+- `ninja` or `samurai`
```
gsliepen, line 9: Maybe we should also test building with samurai?

INSTALL.md
```diff
@@ -1,14 +1,27 @@
...
 - `pkgconf` or `pkg-config`
 - `GCC` or `Clang` (any version with C11 support, although older versions might work)
 - `OpenSSL`\* (1.1.0+) or `LibreSSL` or `libgcrypt` (not needed if legacy protocol is disabled)
 
+### No Python?
+
+If you're on a constrained system that doesn't have (or cannot run) Python, you can try building tinc with [muon][muon],
+which is a pure C reimplementation of the same idea.
+Please note that `meson` is considered to be the main way of building tinc, and `muon` is supported on a best-effort basis.
+Some optional features or tests might not work.
```
gsliepen, line 19: I don't think we have any features in tinc itself that wouldn't work when building with muon. The only issue is the test suite.

gsliepen (changes requested):
For me it fails to build with muon due to this error:
```
[...]
warn dependency lzo2 not found
info found library 'lzo2' at '/usr/lib/x86_64-linux-gnu/liblzo2.so'
/home/guus/scratch/tinc-1.1/src/meson.build:325:16: error method external_library.type_name() not found
325 |     if dep_lzo.type_name() == 'internal' or cc.has_header('lzo/lzo1x.h', dependencies: dep_lzo)
                     ^
/home/guus/scratch/tinc-1.1/meson.build:148:1: error in function subdir()
148 | subdir('src')
      ^
```

INSTALL.md
hg: It's already used on alpine (where it's aliased to ninja), but I'll add to to the muon job anyway.

gsliepen:
> For me it fails to build with muon

Ah, I did not have libpkgconf-dev installed when building muon. With that in place, it works fine.

hg:
> Ah, I did not have libpkgconf-dev installed when building muon. With that in place, it works fine.

I guess wrapped lzo will not be available until `type_name()` is implemented in muon.

It was actually the reason for that clause about "optional features" not working.

IIRC `cc.has_header()` doesn't look up headers in subprojects, so that clause cannot be removed if we want to have best experience for meson users.

---


samurai embedding seems buggy. The `.so` is not installed automatically, you have to copy it and then do some ld configuration, and even after that it failed on my machine with weird errors.

It also requires ninja or samurai to be available, which kind of defeats the point.

So we build samurai from source, too.

.ci/compat/run.sh
```diff
@@ -0,0 +1,251 @@
...
+  set Port 30002
+  set Interface tinc10
+  set Subnet ${addr[tinc10]}
+  set Address localhost
+  set Compression 10
+  set LogLevel 5
+EOF
+
+# Remove EC public key or newer nodes will complain about protocol downgrade
+sed -i 's#^Ed25519PublicKey##' $etc/tinc10/hosts/tinc10
```
gsliepen, line 190: You could just add `del Ed25519PublicKey` to the commands you pass to the CLI. But more importantly, it might be better to initialize an 1.0 node the 1.0 way, by creating a `tinc.conf` and `hosts/tinc10` file with everything but the key, and then run `tincd -c $etc/tinc10 -K` to generate the key.

.ci/muon/run.sh
```diff
@@ -0,0 +1,58 @@
...
+make install
+popd
+
+header 'Fetch and build muon'
+
+git clone --depth=1 $git_muon ~/muon
+pushd ~/muon
+./bootstrap.sh build
+./build/muon setup build
+samu -C build
```
gsliepen, line 38: Ah but if samu is built in to muon, you need to do this:
```
./build/muon -C build samu
```
No need to build and install samurai separately this way.

.ci/compat/run.sh
hg: The thought was to check that newer CLIs can create something that works even with old servers, but this makes more sense. Replaced.

.ci/muon/run.sh
hg: I think the current approach is not only simpler, but also more realistic.

You need ninja or samurai to build muon, so there's the chicken and egg problem.

And I had several issues with using samurai through muon:

1. `samu.so` isn't installed automatically, so you have to copy it manually and use `LD_LIBRARY_PATH` or add `/usr/local/lib` to dynamic linker lookup paths
1. then it failed to recompile a build directory which both the original separate samu and ninja worked fine with both before and after (just errors out with an ok message, but non-zero exit code).

The first one could probably be solved by static linking, but we run into more issues since `libcurl-dev` doesn't support it (and it's necessary to fetch wraps).

The second is more problematic. I'll see if it's reproducible, and report to upstream if it is.

I was very impressed with muon as a whole, but this particular feature leaves a taste of WIP.

gsliepen: Ok, let's keep it like this then. I also think muon is already pretty decent; we don't have a trivial meson setup and it handles almost everything fine.

eli-schwartz: The embedded samu is considered rather experimental and hacky by the muon author, that being said, it should be getting built as a static library and statically linked into muon itself -- the samu subproject defines `default_options: ['default_library=static']`. So there should be no .so to install.

eli-schwartz: Anyway, having a standalone samu is not unreasonable... :D

PR #404, opened 2022-06-02
Add more integration tests

hg:
A few more tests mostly related to device configuration. These will fail without #403.

https://github.com/hg/tinc/actions/runs/2428377765

I think all of the easy pickings are covered. The remaining variables are used deep in tincd internals and require lots of setup to test (often needing root and functioning network, and thus only working on Linux). I'll continue adding tests for them, but more slowly because of this.

### DIFF
--- a/test/integration/device_multicast.py
+++ b/test/integration/device_multicast.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python3
+
+"""Test multicast device."""
+
+import os
+import socket
+import struct
+import time
+
+from testlib import check
+from testlib.log import log
+from testlib.proc import Tinc, Script
+from testlib.test import Test
+
+MCAST_ADDR = "224.15.98.12"
+PORT = 38245
+
+
+def multicast_works() -> bool:
+    """Check if multicast is supported and works."""
+
+    msg = b"foobar"
+
+    try:
+        with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as server:
+            server.bind((MCAST_ADDR, PORT))
+
+            req = struct.pack("=4sl", socket.inet_aton(MCAST_ADDR), socket.INADDR_ANY)
+            server.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP, req)
+
+            with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as client:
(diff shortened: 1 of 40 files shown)


test/integration/device_multicast.py
```diff
@@ -0,0 +1,83 @@
...
+from testlib.log import log
+from testlib.proc import Tinc, Script
+from testlib.test import Test
+
+
+def have_multicast() -> bool:
+    """Can we run multicast tests?"""
+
+    if platform.system() != "Linux":
+        return False
```
gsliepen, line 22: Hm, multicast should work everywhere. If you want to test if the operating system it's running on support multicast, maybe just create two multicast enabled sockets in Python and check if they can send/receive to each other?

test/integration/bind.py
```diff
@@ -0,0 +1,94 @@
...
+
+ADDRESSES = ("127.253.85.1", "127.253.85.2", "127.253.85.3")
+
+
+def get_listen_sockets(proto: str) -> str:
+    """Get a list of the currently listening sockets (ss output)."""
+    listen = subp.run(
+        ["ss", f"-nl{proto[0]}"], check=True, stdout=subp.PIPE, encoding="utf-8"
+    )
+    return listen.stdout
```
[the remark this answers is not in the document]
hg: Split into two tests as you suggested. One of them is still using `ss` because we need some way to check that UDP sockets are listening. Since we need to figure out which interfaces and addresses are available, and there's no way to do this in a cross-platform way, the test is tied to Linux and `ip addr` in any case.

TCP is checked with Python's `socket`.

test/integration/device_multicast.py
hg: Thanks, that was pretty stupid of me. As to why this check is needed at all, Windows doesn't support multicast without (it seems) additional configuration. I've also run into this on Linux (can't remember the exact setup, probably a `dummy` device with broadcast disabled and `lo` down). So it is possible.

gsliepen: You're right, if we can't guarantee the network setup (it could even be firewall rules blocking multicast packets), then this test might be problematic. Maybe it should only be run in a network namespace then...

gsliepen: But it looks good enough.

Issue #255, opened 2020-12-10
Client hangs on invalid invitation


haering:
I'm using tinc 1.1pre17.

When trying to connect to a server with a invalid invitation (maybe used before or completly invalid) the server denies the invitation but the client hangs on connecting.

Server log:
```
VPN     | Connection from client port 59781
VPN     | Peer client port 59781 tried to use non-existing invitation [Invitationcode]
VPN     | Closing connection with <unknown> (client port 59781)
```
Client log:
```
.\tinc.exe  join server:655/[Invitationcode]
Connected to server port 655...
```
The expected behaviour would the client exiting with an error code instead of waiting.

PR #405, opened 2022-06-04
Notify 'tinc join' about invitation errors

hg:
See #255

# New server, new client

```
Connected to localhost port 13394...
Request rejected: invitation is invalid or missing.
Invitation cancelled.
(exit code 1)
```

# New server, old client

```
Connected to localhost port 13394...
Invitation cancelled.
(exit code 1)
```

# Old server, new client

```
Connected to localhost port 655...
(10-second pause)
Timed out waiting for server to reply.
(exit code 1)
```

I'll refactor sptps record codes (`1`/`2`/`3`) in a separate PR.

### DIFF
--- a/src/invitation.c
+++ b/src/invitation.c
@@ -1196,6 +1196,21 @@ static bool invitation_receive(void *handle, uint8_t type, const void *msg, uint
 	return true;
 }
 
+bool wait_socket_recv(int fd) {
+	fd_set fds;
+	FD_ZERO(&fds);
+	FD_SET(fd, &fds);
+
+	struct timeval tv = {.tv_sec = 5};
+
+	if(select(fd + 1, &fds, NULL, NULL, &tv) != 1) {
+		fprintf(stderr, "Timed out waiting for the server to reply.\n");
+		return false;
+	}
+
+	return true;
+}
+
 int cmd_join(int argc, char *argv[]) {
 	free(data);
 	data = NULL;
@@ -1320,6 +1335,7 @@ next:
 			freeaddrinfo(ai);
 			free(b64_pubkey);
 			ecdsa_free(key);
+			fprintf(stderr, "Could not connect to inviter. Please make sure the URL you entered is valid.\n");
 			return 1;
 		}
 	}
@@ -1383,7 +1399,7 @@ next:
 	}
 
 	if(memcmp(hishash, hash, 18)) {
-		fprintf(stderr, "Peer has an invalid key!\n%s\n", line + 2);
+		fprintf(stderr, "Peer has an invalid key. Please make sure you're using the correct URL.\n%s\n", line + 2);
 		ecdsa_free(key);
 		return 1;
 
@@ -1409,7 +1425,7 @@ next:
 		goto exit;
 	}
 
-	while((len = recv(sock, line, sizeof(line), 0))) {
+	while(wait_socket_recv(sock) && (len = recv(sock, line, sizeof(line), 0))) {
 		if(len < 0) {
 			if(sockwouldblock(sockerrno)) {
 				continue;
@@ -1452,7 +1468,7 @@ exit:
 	closesocket(sock);
 
 	if(!success) {
-		fprintf(stderr, "Invitation cancelled.\n");
+		fprintf(stderr, "Invitation cancelled. Please try again and contact the inviter for assistance if this error persists.\n");
 		return 1;
 	}
 
--- a/doc/tinc.conf.5.in
+++ b/doc/tinc.conf.5.in
@@ -384,7 +384,7 @@ This only has effect when
 .Va Mode
 is set to
 .Qq switch .
-.It Va MaxConnectionBurst Li = Ar count Pq 100
+.It Va MaxConnectionBurst Li = Ar count Pq 10
 This option controls how many connections tinc accepts in quick succession.
 If there are more connections than the given number in a short time interval,
 tinc will reduce the number of accepted connections to only one per second,
--- a/doc/tinc.texi
+++ b/doc/tinc.texi
@@ -1153,7 +1153,7 @@ This option controls the amount of time MAC addresses are kept before they are r
 This only has effect when Mode is set to @samp{switch}.
 
 @cindex MaxConnectionBurst
-@item MaxConnectionBurst = <@var{count}> (100)
+@item MaxConnectionBurst = <@var{count}> (10)
 This option controls how many connections tinc accepts in quick succession.
 If there are more connections than the given number in a short time interval,
 tinc will reduce the number of accepted connections to only one per second,
--- a/src/invitation.h
+++ b/src/invitation.h
@@ -23,4 +23,8 @@
 int cmd_invite(int argc, char *argv[]);
 int cmd_join(int argc, char *argv[]);
 
+// Wait until data can be read from socket, or a timeout occurs.
+// true if socket is ready, false on timeout.
+bool wait_socket_recv(int fd);
+
 #endif
--- a/src/tincctl.c
+++ b/src/tincctl.c
@@ -495,6 +495,10 @@ bool recvline(int fd, char *line, size_t len) {
 	}
 
 	while(!(newline = memchr(buffer, '\n', blen))) {
+		if(!wait_socket_recv(fd)) {
+			return false;
+		}
+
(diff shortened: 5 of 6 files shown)


src/protocol_auth.c
```diff
@@ -189,7 +193,7 @@ static bool receive_invitation_sptps(void *handle, uint8_t type, const void *dat
 	}
 
 	if(type != 0 || len != 18 || c->status.invitation_used) {
-		return false;
+		return notify_invitation_error(c, "bad request");
```
gsliepen, line 196: This changes the return value from `false` to `true`, meaning the connection will stay open. We want it to be closed right after sending the error message.

hg: How about using a separate status flag in `connection_t`? I don't think there's a way to flush the output buffer *right now* without blocking the thread for an indeterminate amount of time (which is why connection termination was left to the client). If we return `false` here, the connection is terminated without sending anything.

gsliepen: Ah yes. The problem is that an attacker might try to join using a false invitation. We don't want them to tie up resources unnecessarily, and we also don't want to give them information that might be used to determine if parts of the join request are valid. So I'm thinking we should not notify tinc join about errors at all. If the connection is closed prematurely, that's the sign that the invitation was not accepted.

hg: Makes sense. I removed tarpit for both phases of the connection (pre & post-sptps) since (as you say) the only thing it does it tie up server resources and confuse users (re #255). Calling tarpit on connection burst seems to take care of script kiddies with dictionary bruteforce without worsening experience for legitimate users.

Thanks for spending time on this, I know you could have added this yourself in about 5 seconds; review takes a lot more effort.

src/invitation.c
```diff
@@ -1196,6 +1196,10 @@ static bool invitation_receive(void *handle, uint8_t type, const void *msg, uint
 	return true;
 }
 
+static void invitation_rejected(void) {
+	fprintf(stderr, "Invitation rejected. Please consult server logs for details.\n");
+}
```
gsliepen, line 1201: You're right that the current (lack of) error message is bad. However, telling the person that does a `tinc join` to check server logs is not very helpful; they might not be the ones running the server. Maybe something like "contact the inviter for more details" might be better.

The other issue is that there are different ways to fail that even the client can distinguish. For example, I see you call this function also in case that it could not resolve the hostname in the invitation URL. That actually means the invitation was never used, so it isn't rejected. Only after sending the server the cookie can the invitation be rejected. If the connection is dropped before that, or if there is some name lookup problem, I think the error message should be "Could not contact the inviter."

Around line 1307, where we have `if (!ai)`, that's where we should explicitly give the error message "Could not resolve address from invitation URL", as `str2addrinfo()` will not print anything itself in this case.

src/net.c
```diff
@@ -303,7 +303,9 @@ static void periodic_handler(void *data) {
 
 void handle_meta_connection_data(connection_t *c) {
 	if(!receive_meta(c)) {
-		if(!c->status.control) {
+		// Terminate invitation connections right away since
+		// hanging connections confuse and frustrate users
+		if(!c->status.control && !c->status.invitation) {
```
gsliepen, line 308: I did say we should not spend resources on botched join connections, but we should tarpit those connections anyway. The tarpit will automatically close connections if there are too many of them, so it only uses a fixed number of filedescriptors. We do want to frustrate attackers who are trying joins.

src/protocol_auth.c
gsliepen: See my comment about tarpitting in the code. I might have done this in 5 seconds, if I've had the idea of improving this. But you are doing a lot of work now that I might not even have attempted, and doing things different that what I would have done sometimes gives me a new perspective. So I'm fine with these reviews taking more effort, it's a net positive IMO :)

hg:
Not sure it's possible to resolve #255 properly if we want to both tarpit suspicious connections, and not leave the user hanging (since every failing connection is treated as suspicious).

I added timeouts to both places where client could hang and didn't touch the server,

<details>

<summary>aside</summary>

although TBH I don't see what prevents a malicious actor from doing the same. Something like Go makes writing these things really easy: start `MaxConnectionBurst` goroutines, split a dictionary between them, and try words in a loop with a 1-second timeout until one of them goes through (which both works around the tarpit, and prevents you from running into connection burst limits).

Or even something like this:

```sh
while read -rp word; do
  url=$(build_url "$word")
  timeout 1 tinc join nsa.gov/"$url" && {
    echo "success: $word"
    exit 0
  }
done
```

Start ten of them in parallel (using the appropriately named GNU `parallel`) until one finds something.

I guess this can be prevented by remembering IPs and adding per-IP limits, but with increasing use of CGNAT it may do more harm than good (a bad guy can buy hundreds of proxies for pennies, or just use tor).

Purely in theory, if connection is closed by the server, the client can guess the reason by accurate timing, but this can be worked around by closing it after a randomly generated delay.

---

</details>

`tinc join` output has been updated to encourage the user to seek assistance if repeated joins fail to work. All possible code paths seem to provide their own error messages already (including `str2addrinfo()`: it lets you know when it fails to look up a domain).

<details>

<summary>A few examples (covered by new tests)</summary>

There are basically three error modes:

- obviously bad URL (we notify the user of that)
- connection error (the problem can be on either side of the connection; we ask to double-check the URL, maybe they copied it wrong)
- server error (probably; we ask the user to retry and contact the guy on the other side if it doesn't help)


## bad domain

```
$ /tmp/build/src/tinc -c i join frobnicator:13394/C92j2YTpTzZoiS3bR_-F873nMr9bDxW5Ut4ONNrhyN8-_10g
Error looking up frobnicator port 13394: Name or service not known
```

## bad url

```
$ /tmp/build/src/tinc -c i join frobnicator
Invalid invitation URL.
```

## bad port

```
$ /tmp/build/src/tinc -c i join localhost:1337/QUAuoWrDjKB6QT1IJhpmydsSCC2N2B3-6NnCeXpEeNZauuif
Could not connect to 127.0.0.1 port 1337: Connection refused
Could not connect to inviter. Please make sure the URL you entered is valid.
```

## URL from another server

```
$ /tmp/build/src/tinc -c i join 'localhost:13394/vaMdJSNRsKa834yYbPCP2JV2oFHTYsKgooYiRJrmV9GRvdf1'
Connected to localhost port 13394...
Peer has an invalid key. Please make sure you're using the correct URL.
9CErgCAUk1R4ugLUxCmODMHbP+682Smt+XxVTTr7joB
```

## broken secret

```
$ /tmp/build/src/tinc -c i join frobnicator:13394/C92j2YTpTzZoiS3bR_-F873nMr9bDxW5Ut4ONNrhyN8-_10
Invalid invitation URL.
```

## almost correct URL (one char replaced)

```
ERROR   Peer 127.0.0.1 port 42572 tried to use non-existing invitation yNC0rjfFMcs-oC1lpHMBCndl
```

```
$ /tmp/build/src/tinc -c i join localhost:13394/QUAuoWrDjKB6QT1IJhpmydsSCC2N2B3-6NnCeXpEeNZauuif
Connected to localhost port 13394...
(5-second pause)
Timed out waiting for the server to reply.
Invitation cancelled. Please try again and contact the inviter for assistance if this error persists.
```

## `invitations` removed from the server

Error before sptps is established

```
ERROR   Got invitation from 127.0.0.1 port 53768 but we don't have an invitation key
```

```
$ /tmp/build/src/tinc -c i join 'localhost:13394/vaMdJSNRsKa834yYbPCP2JV2oFHTYsKgooYiRJrmV9GRvdf1'
Connected to localhost port 13394...
(5-second pause)
Timed out waiting for the server to reply.
Invitation cancelled. Please try again and contact the inviter for assistance if this error persists.
```

## bad URL

Error after sptps is established

```
ERROR   Peer 127.0.0.1 port 38490 tried to use non-existing invitation DYihGIhImzevayj1U1WCBA_c
```

```
$ /tmp/build/src/tinc -c i join localhost:13394/C92j2YTpTzZoiS3bR_-F873nMr9bDxW5Ut4ONNrhyN8-_10g
Connected to localhost port 13394...
(5-second pause)
Timed out waiting for the server to reply.
Invitation cancelled. Please try again and contact the inviter for assistance if this error persists.
```

</details>

gsliepen:
Tarpitting is not a perfect defense at all, but it does help against low volume attacks. Having timeouts in the client is a good thing to add, even if we didn't tarpit. The new error messages are a huge improvement!

PR #445, opened 2023-12-12
Add distribution specific package installation instructions

marek22k:


INSTALL.md
gsliepen, line 1: When rendered to HTML, these lines will just form one paragraph. To ensure they stay separate lines, make them bullet lists. I would also just have one list of installation instructions that include all packages, as the distributions carry all dependencies.

gsliepen (changes requested):

INSTALL.md
marek22k: Thanks for the hints! I have now placed the installation of all dependencies under the new item "Linux".

gsliepen (approved):

gsliepen:
Merged in commit a3328b5886bb13f582bf23631761d348f4dba26d. Thanks!

PR #446, opened 2023-12-12
Correct a type mismatch

marek22k:
In one place in the source code, the debug level of type `int` was used and in other places of type `debug_t`. This commit introduces a change so that the type `debug_t` is used in both places. This also resolves a compiler warning.

### DIFF
--- a/src/info.c
+++ b/src/info.c
@@ -26,7 +26,7 @@
 #include "tincctl.h"
 #include "utils.h"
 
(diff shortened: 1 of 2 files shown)


gsliepen (approved):

gsliepen:
Merged in commit 6e6dcbc7e77fbc4f7873b6e40c911b7038a6a648. Thanks for the fix!

Issue #447, opened 2024-01-22
[Bug] Adding a simple failing test case for sssp_bfs()


crazyboycjr:
The following testcase fails. I found it because I am experiencing this exact issue in my personal setup. The problem is due to an incorrect implementation here https://github.com/gsliepen/tinc/blob/1.1/src/graph.c#L187. In short, what to compare should be `dist[u] >?= dist[v] + g[u][v]`. The existing code only compares the weight of the last edge.

```C
static void test_sssp_bfs_2(void **state) {
	(void)state;

	node_t *mars = make_node("mars");
	node_t *saturn = make_node("saturn");
	node_t *neptune = make_node("neptune");

	//          1000            500
	// myself ---------- mars ------------- neptune
	//      \                              /
	//       ------- saturn --------------
	//          50               501

	// Upper route
	connect_nodes(myself, mars, 1000);
	connect_nodes(mars, neptune, 500);

	// Lower route
	connect_nodes(myself, saturn, 50);
	connect_nodes(saturn, neptune, 501);

	sssp_bfs();

	assert_true(mars->status.visited);
	assert_true(saturn->status.visited);
	assert_true(neptune->status.visited);

	assert_false(mars->status.indirect);
	assert_false(saturn->status.indirect);
	assert_false(neptune->status.indirect);

	assert_int_equal(1, mars->distance);
	assert_int_equal(1, saturn->distance);
	assert_int_equal(2, neptune->distance);

	assert_ptr_equal(mars, mars->nexthop);
	assert_ptr_equal(saturn, saturn->nexthop);
	assert_ptr_equal(saturn, neptune->nexthop);

	assert_ptr_equal(lookup_edge(myself, mars), mars->prevedge);
	assert_ptr_equal(lookup_edge(myself, saturn), saturn->prevedge);
	assert_ptr_equal(lookup_edge(saturn, neptune), neptune->prevedge);

	assert_ptr_equal(mars, mars->via);
	assert_ptr_equal(saturn, saturn->via);
	assert_ptr_equal(neptune, neptune->via);
}
```

@gsliepen @hg This is simply adding a testcase. Could you please take a look?

crazyboycjr:
I have also created a fix for that. The repo seems not having been actively accepting PR recently, so I just post the patch here. @gsliepen let me know if you would like me to send a PR.

```C
diff --git a/src/graph.c b/src/graph.c
index 5a7a16a7..5bec2c16 100644
--- a/src/graph.c
[...]

PR #449, opened 2024-01-22
Fix bug in shortest path implementation

crazyboycjr:
#447

This PR should fix the algorithm implementation in shortest path. The unit test created in #448 should pass now.


crazyboycjr:
@gsliepen The patch did not touch other part of the code base. The failing CI jobs should be irrelevant. @gsliepen Could you please take a look?

PR #485, opened 2026-04-03
fix: resolve -Wdiscarded-qualifiers warnings in ifconfig_route and complete_config

heitbaum:
In ifconfig_route, remove const from the value parameter as it is written through via sep (*sep++ = 0) and is only called from finalize_join() with a mutable buffer.

In complete_config, change dot to const char* as it is only used for pointer arithmetic and reading, never written through.

### DIFF
--- a/src/ifconfig.c
+++ b/src/ifconfig.c
@@ -178,7 +178,7 @@ void ifconfig_address(FILE *out, const char *value) {
 #endif
 }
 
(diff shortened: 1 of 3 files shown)


gsliepen:
Cherry-picked as commit 3ae8fd111bc2390417661e09c18a6aec4d5e729f. Thanks!

